PR #36, opened 2026-02-26
Revise Internet Identity SKILL documentation

aterga:
Updated the Internet Identity authentication documentation to reflect changes in dependencies, versioning, and descriptions. Improved clarity on local development setup and common mistakes.


marc0olo (approved):
approving this for now. we will continously improve.

it feels like we have a bit too much unnecessary installation instructions, but I want to tackle to frontmatter refactoring first before applying other changes.

PR #58, opened 2026-03-02
feat: add canister-security skill

marc0olo:
## Summary

- Adds `canister-security` skill covering 12 IC-specific security pitfalls with links to official DFINITY security best practices
- Complete Motoko (core 2.0) and Rust (ic-cdk 0.19) implementations for access control, saga pattern, and inspect_message
- Adds `@dfinity/security` as CODEOWNER for `skills/canister-security/`
- Adds cross-referencing guidance to CONTRIBUTING.md — domain skills should reference `canister-security` rather than inlining general security patterns

**Relationship to #57:** Complementary. #57 adds multi-canister-specific security content (reentrancy in factory patterns, production readiness). This skill covers general IC security fundamentals that apply to all canisters. Both should coexist.


skills/canister-security/SKILL.md
```diff
@@ -0,0 +1,335 @@
...
+```motoko
+import Principal "mo:core/Principal";
+import Array "mo:core/Array";
+import Runtime "mo:core/Runtime";
+
+persistent actor {
+
+  // --- Authorization state ---
+  stable var owner : Principal = Principal.fromText("aaaaa-aa"); // replaced during init
+  stable var admins : [Principal] = [];
```
venkkatesh-sekar:
```suggestion
 var owner : Principal = Principal.fromText("aaaaa-aa"); // replaced during init
 var admins : [Principal] = [];
```

venkkatesh-sekar (changes requested):
@marc0olo I took a first pass and left feedback. But the security team is also doing some parallel work here https://github.com/dfinity-lab/security-skills, so maybe we can unify the effort.

skills/canister-security/SKILL.md
```diff
@@ -0,0 +1,454 @@
...
+    let balance = get_balance(&caller);
+    if balance < amount {
+        return Err("insufficient balance".to_string());
+    }
+    deduct_balance(&caller, amount);
+
+    // 3. Compensation guard — credits back on drop unless completed.
+    //    Drop runs during cleanup even if callback traps, so compensation persists.
+    //    Putting credit_balance only in the Err arm would be rolled back on trap.
+    let mut compensation = CompensationGuard::new(caller, amount);
```
venkkatesh-sekar: Same recommendation, please use only the CallerGuard example as the CompensationGuard is a complex topic.

For example, in the happy `Ok()` path, if there is a trap and Drop is run, complete will be false and the balance will be refunded irrespective of the intercanister call status. 

bounded_wait call also returns an error on Timeout, in this case blindly refunding is wrong since the balance may or may not have been transferred.

raymondk (approved):

PR #60, opened 2026-03-03
Revise HTTPS Outcalls skill documentation

fspreiss:
Updates the HTTP Outcall skill documentation as follows:

* **Fix Motoko example to compile and run correctly.** Replace hand-rolled type definitions with the compiler-provided import IC "ic:aaaaa-aa" types, add the missing is_replicated field, and use Blob directly instead of [Nat8] for the response body.

* **Fix consensus failures in POST examples.** Add a dedicated transformPost function (both Motoko and Rust) that strips the response body in addition to headers, because httpbin.org's /post endpoint returns the caller's IP in the origin field, which differs across replicas. POST functions now return a status message instead of the non-deterministic body.

* **Use automatic cycle attachment.** Replace hardcoded cycle amounts with Call.httpRequest from the ic mops package (Motoko) and document that ic_cdk::management_canister::http_request (Rust) does the same. Both use the ic0.cost_http_request system API under the hood.

* **Add pitfall for omitting max_response_bytes.** New pitfall #5 warns that omitting it defaults to 2MB and charges ~21.5B cycles on a 13-node subnet.

* **Correct the cycle cost formula.** The original formula had request_bytes and response_bytes swapped. Updated with the correct per-byte rates (5,200/request byte, 10,400/max_response_bytes byte on a 13-node subnet), explained that the charge is against max_response_bytes (not actual response size), and documented the ic0.cost_http_request system API for runtime cost calculation.

* **Update prerequisites** to require moc >= 0.14.10 and ic >= 2.1.0 in mops.toml.

### DIFF
--- a/skills/multi-canister/SKILL.md
+++ b/skills/multi-canister/SKILL.md
@@ -39,21 +39,21 @@ Calls may be unbounded wait (caller MUST wait until the callee produces a respon
 
 **When NOT to use:** Simple apps with <1GB data. Single-canister is simpler, faster, and avoids inter-canister call overhead. Do not over-architect.
 
-## Mistakes That Break Your Build
+## Issues that May Cause Functional Bugs
+
+For building a multi-canister application, take the perspective of an experienced senior software engineer and carefully read the following issues that may cause subtle functional bugs. Meticulously avoid bugs that could be caused by these issues. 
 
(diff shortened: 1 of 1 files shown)


derlerd-dfinity (approved):
Nice! Thanks a lot!

PR #80, opened 2026-03-06
feat: deploy site via IC asset canister at skills.internetcomputer.org

marc0olo:
## Summary

- Deploy the site to an IC asset canister instead of GitHub Pages, serving at `skills.internetcomputer.org`
- Add `icp.yaml` config, `.ic-assets.json5` with explicit `charset=utf-8` headers, and IC deploy workflow with secure identity handling (base64 via env var, scoped to `IC mainnet` environment)
- Remove the GitHub Pages deploy workflow (`deploy.yml`)
- Make `SITE_URL` configurable via env var; interactive components (skill cards, access page) use `window.location.origin` so the site works on any domain
- Update `llms.txt` / `llms-full.txt` to use site URLs and add Skills Discovery RFC section
- Update all hardcoded `dfinity.github.io/icskills` references to `skills.internetcomputer.org`
- Update how-it-works terminal to showcase Agent Skills Discovery RFC (`.well-known/skills/index.json`)

### DIFF
--- a/public/.well-known/ic-domains
+++ b/public/.well-known/ic-domains
@@ -1,2 +1,1 @@
-skills.icp.build
 skills.internetcomputer.org


public/.well-known/ic-domains
```diff
@@ -0,0 +1,2 @@
+skills.icp.build
```
raymondk: Maybe we don't host on this since the generated metadata hard codes the ic.org host

raymondk (approved):

PR #158, opened 2026-04-09
feat: Add encrypted-maps skill for vetKeys EncryptedMaps

andreacerulli:
New skill covering end-to-end encrypted on-chain storage using the ic-vetkeys EncryptedMaps library. Includes complete Rust backend, Motoko backend reference, frontend TypeScript examples, icp-cli configuration, and pitfalls from real-world testing.

### DIFF
--- a/skills/encrypted-maps/SKILL.md
+++ b/skills/encrypted-maps/SKILL.md
@@ -346,16 +346,14 @@ A complete Motoko backend is available in `references/motoko-backend.md`. It exp
 
 ### icp.yaml
 
-**Rust:**
-
 ```yaml
 canisters:
   - name: backend
     recipe:
-      type: "@dfinity/rust@v3.2.0"
+      type: "@dfinity/rust@v3.2.0"       # Motoko: "@dfinity/motoko@v4.1.0"
       configuration:
-        package: encrypted-maps-backend
-    init_args: '("test_key_1")'
+        package: encrypted-maps-backend   # Motoko: main: backend/src/Main.mo
+    init_args: '("test_key_1")'           # Use '("key_1")' for production
 
   - name: frontend
     recipe:
@@ -372,33 +370,7 @@ networks:
     ii: true
 ```
 
-**Motoko:**
-
-```yaml
-canisters:
-  - name: backend
-    recipe:
-      type: "@dfinity/motoko@v4.1.0"
-      configuration:
-        main: backend/src/Main.mo
-    init_args: '("test_key_1")'
-
-  - name: frontend
-    recipe:
-      type: "@dfinity/asset-canister@v2.1.0"
-      configuration:
-        dir: frontend/dist
-        build:
-          - npm --prefix frontend install
-          - npm --prefix frontend run build
-
-networks:
-  - name: local
-    mode: managed
-    ii: true
-```
-
-Change `init_args` to `'("key_1")'` for production. See the `icp-cli` skill for the full configuration reference.
+See the `icp-cli` skill for the full configuration reference.
 
 ### Frontend (TypeScript)
 
@@ -450,12 +422,6 @@ await encryptedMaps.setValue(owner, mapName, mapKey, value);
 const decrypted = await encryptedMaps.getValue(owner, mapName, mapKey);
 const data = JSON.parse(new TextDecoder().decode(decrypted));
 
-// Update — same as store, overwrites existing value
-const updated = new TextEncoder().encode(
-    JSON.stringify({ username: "me@gmail.com", password: "n3w_s3cret" }),
-);
-await encryptedMaps.setValue(owner, mapName, mapKey, updated);
-
 // Delete a single value
 await encryptedMaps.removeEncryptedValue(owner, mapName, mapKey);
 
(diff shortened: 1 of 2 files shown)


skills/encrypted-maps/SKILL.md
```diff
@@ -0,0 +1,492 @@
...
+    "@dfinity/agent": "^3.4.0",
+    "@dfinity/principal": "^3.4.0",
+    "@dfinity/vetkeys": "^0.4.0",
+    "@icp-sdk/auth": "^5.0.0",
+    "@icp-sdk/core": "^5.0.0"
+  }
+}
+```
+
+`@dfinity/agent` and `@dfinity/principal` are required because `@dfinity/vetkeys` depends on them. `@icp-sdk/auth` provides Internet Identity login — import from `@icp-sdk/auth/client`, not the root module (the root has no exports). `@icp-sdk/core` provides canister ID discovery via the `ic_env` cookie set by `icp deploy`.
```
marc0olo, line 83: same question here. can we get rid of using `@dfinity/agent` and `@dfinity/principal` ?

marc0olo:
## Review: encrypted-maps skill

Strongest of the two PRs — the structure is clear, the pitfalls section is excellent, and the frontend is well-organized into logical sub-sections. A few things to fix before merge:

### Bug: `remove_map_values` return type mismatch

The function signature declares `Result<Vec<EncryptedMapValue>, String>` but the closure maps each removed entry to `ByteBuf::from(...)`, so the inferred return type is `Result<Vec<ByteBuf>, String>`. These don't match — the code won't compile as-is. Fix the return type to match the closure.

### Pitfall 4 — cycle cost is vague

*"The library attaches cycles automatically, but the canister must have sufficient balance. Top up before testing."* — an agent has no idea how much to top up. Add the exact cost: each `get_encrypted_vetkey` call costs `26_153_846_153` cycles (both locally on PocketIC and on mainnet, for both `test_key_1` and `key_1`). The same PocketIC deviation note from the `vetkeys` skill applies here: locally, `test_key_1` costs the same as `key_1`, not less.

### Deploy section — explain the cycle top-up amount

`icp cycles top-up backend 10T -e ic` is reasonable but unexplained. Add a brief note: each `get_encrypted_vetkey` call costs ~26B cycles, so 10T covers roughly 380 key derivations as an initial buffer.

### Missing: `#[post_upgrade]` note

Since EncryptedMaps uses stable memory exclusively, upgrades work without an explicit `#[post_upgrade]` handler — state survives automatically. One sentence clarifying this prevents agents from adding an unnecessary handler or worrying about data loss on upgrade.

### Description — make it the explicit default

Consider adding that `encrypted-maps` is the **default starting point** for any encrypted storage need, and `vetkeys` is only for when lower-level control is required. Something like: *"Start here for any encrypted storage use case; escalate to `vetkeys` only if you need BLS signatures, custom IBE, or timelock."* This helps agents make the right choice on ambiguous queries like "I want to add encryption to my app."

### `icp.yaml` section

The `icp-cli` skill covers recipes, frontend setup, and network config. Collapse this section to the encrypted-maps-specific part only:

```yaml
canisters:
  - name: backend
    init_args: '("test_key_1")'  # Production: '("key_1")'
```

With a reference: *"For full icp.yaml setup, see the `icp-cli` skill."*

### Deploy & Verify

Drop the generic `icp network start -d` and `icp deploy backend` — those are `icp-cli` territory. The domain-specific verification calls are good — keep those. Add a note that `get_vetkey_verification_key` triggers an actual management canister call and costs cycles, so the canister needs to be topped up first.

### Missing: evaluation file

Please add `evaluations/encrypted-maps.json`. Suggested content:

```json
{
  "skill": "encrypted-maps",
  "description": "Tests whether agents use the EncryptedMaps high-level library correctly, avoid raw vetkeys wiring, handle AccessRights as Candid variants, and understand the 32-byte key/name limit.",

  "output_evals": [
    {
      "name": "Password manager scaffold",
      "prompt": "I want to build a password manager on ICP where each user's passwords are encrypted and only they can read them. Show me just the frontend store and retrieve flow, no backend or deploy steps.",
      "expected_behaviors": [
        "Uses EncryptedMaps class from @dfinity/vetkeys/encrypted_maps",
        "Calls setValue() to store and getValue() to retrieve — does NOT call vetkd_derive_key directly",
        "Map name and map key are encoded with TextEncoder and are within 32 bytes",
        "Does NOT manually generate transport keys, wire AES-GCM, or call management canister methods"
      ]
    },
    {
      "name": "Multi-user vault sharing",
      "prompt": "How do I let a vault owner share read access to their encrypted map with another user? Show just the TypeScript sharing and revocation calls.",
      "expected_behaviors": [
        "Uses encryptedMaps.setUserRights() with { ReadWrite: null } or { Read: null } — NOT a string",
        "Uses encryptedMaps.removeUser() to revoke",
        "Passes the owner's principal as the first argument to setUserRights",
        "Does NOT re-encrypt the data for the new user — shared encryption handles this automatically"
      ]
    },
    {
      "name": "Adversarial: AccessRights comparison",
      "prompt": "I'm checking user access with `if (rights === 'ReadWrite')` but it's always false even after granting ReadWrite. What's wrong?",
      "expected_behaviors": [
        "Explains that Candid variants are objects, not strings: { ReadWrite: null }",
        "Shows correct comparison: 'ReadWrite' in rights, or rights?.ReadWrite !== undefined",
        "Does NOT suggest using === 'ReadWrite'"
      ]
    },
    {
      "name": "Adversarial: raw vetkeys for storage",
      "prompt": "I want to encrypt user data on ICP. Should I call vetkd_derive_key directly and manage my own AES keys?",
      "expected_behaviors": [
        "Redirects to EncryptedMaps — explains the library handles transport keys, AES-GCM, and key caching",
        "Notes that raw vetkd_derive_key is only needed for BLS, IBE, or timelock (use vetkeys skill)",
        "Does NOT show manual AES-GCM wiring"
      ]
    }
  ],

  "trigger_evals": {
    "description": "Queries to test whether the encrypted-maps skill activates correctly.",
    "should_trigger": [
      "Build a password manager on ICP",
      "I need encrypted notes that only the user can read",
      "How do I store encrypted data on-chain with access control?",
      "I want users to share encrypted files on the Internet Computer",
      "How do I use EncryptedMaps?",
      "Build a secret vault with multi-user sharing on ICP"
    ],
    "should_not_trigger": [
      "I want threshold BLS signatures in my canister",
      "How do I implement identity-based encryption?",
      "Build a sealed-bid auction on ICP",
      "Add Internet Identity login to my dapp",
      "How does stable memory work?"
    ]
  }
}
```

marc0olo:
Superseded by #351, merged as `skills/encrypted-maps/`.

@andreacerulli — your split (high-level encrypted storage separate from the low-level BLS/IBE/timelock primitives) is exactly the architecture that shipped, and the review feedback on this PR fed directly into it. Thank you; you're credited in #351.

What changed against this branch:

- `@icp-sdk/vetkeys` 0.5 (renamed from `@dfinity/vetkeys`, which is frozen at 0.4); frontend agent/identity from `@icp-sdk/core` rather than `@dfinity/agent` / `@dfinity/principal` — the point raised in the inline review here
- Rust `ic-vetkeys` 0.9 + `ic-cdk` 0.20; the backend is now the `export_encrypted_maps_canister!` macro
- Motoko `ic-vetkeys` 0.6 `EncryptedMapsCanister` mixin instead of a hand-rolled backend
- Corrected cycle costs — `test_key_1` and `key_1` behave the same locally and on mainnet
- The metadata / custom-value-endpoint variant lives in `references/metadata.md`
- `evaluations/encrypted-maps.json` shipped with it (4 output cases + bidirectional trigger routing against `vetkeys`)

Closing as superseded.

Issue #156, opened 2026-04-07
vetkd skill: frontend TypeScript examples are wrong for @dfinity/vetkeys 0.4.0; test_key_1 cycle cost incorrect


marc0olo:
## Summary

We built a production-grade vetKeys app from scratch using the ICP skill files as the primary reference — a decentralized family password manager called **KeyChain** ([marc0olo/icp-skill-password-manager](https://github.com/marc0olo/icp-skill-password-manager)). During development, we hit several build-breaking and runtime errors that were **directly caused by incorrect information in the `vetkd` skill**. A secondary gap exists in the `icp-cli` binding-generation reference.

The errors below were verified by fetching the live skill files, not from memory.

---

## 1. `vetkd` skill — Frontend TypeScript API is wrong for `@dfinity/vetkeys` 0.4.0

The skill pins `@dfinity/vetkeys` at **v0.4.0** in the prerequisites but then documents an API that does not exist in that version. All three errors below occur together when following the skill's TypeScript example.

### 1a. `TransportSecretKey.fromSeed()` does not exist

**What the skill says:**
```typescript
const seed = crypto.getRandomValues(new Uint8Array(32));
const transportSecretKey = TransportSecretKey.fromSeed(seed);
const transportPublicKey = transportSecretKey.publicKey();
```

**What `@dfinity/vetkeys` 0.4.0 actually provides:**
```typescript
const transportSecretKey = TransportSecretKey.random();
const transportPublicKey = transportSecretKey.publicKeyBytes();
```

Both `fromSeed()` and `publicKey()` are absent from the 0.4.0 package. Calling them throws a runtime `TypeError: TransportSecretKey.fromSeed is not a function`.

### 1b. `vetKey.toDerivedKeyMaterial()` does not exist

**What the skill says:**
```typescript
const aesKeyMaterial = vetKey.toDerivedKeyMaterial();
const aesKey = await crypto.subtle.importKey(
  "raw",
  aesKeyMaterial.data.slice(0, 32), // 256-bit AES key
  { name: "AES-GCM" },
  false,
  ["encrypt", "decrypt"],
);
```

**What 0.4.0 actually provides:**
```typescript
// asDerivedKeyMaterial is async and returns DerivedKeyMaterial directly
const keyMaterial = await vetKey.asDerivedKeyMaterial();

// DerivedKeyMaterial has built-in encrypt/decrypt — no manual importKey needed
[...]

PR #162, opened 2026-04-10
Replace vetkd skill with a new vetkeys covering BLS, IBE, and timelock.

andreacerulli:
The original vetkd skill was overly low-level and unnecessarily complex. In contrast, the new vetkeys skill focuses on three practical, high-impact primitives: BLS threshold signatures, identity-based encryption (IBE), and timelock encryption.

I evaluated the skill by building example dApps for all three primitives, and the results seemed reasonable. I also compared the experience of developing a Random Beacon dApp with and without the skill. Using the skill produced a more advanced implementation with noticeably less effort. Without it, development took significantly longer, the architecture was more cumbersome, and the end result felt comparatively basic.


skills/vetkeys/SKILL.md
```diff
@@ -0,0 +1,508 @@
...
+serde_cbor = "0.11"
+```
+
+### Frontend
+
+```json
+{
+  "dependencies": {
+    "@dfinity/agent": "^3.4.0",
+    "@dfinity/principal": "^3.4.0",
```
marc0olo, line 51: is there a reason we cannot use the new core library? can we make a new release for the vetkeys lib if it still relies on old dependencies?

andreacerulli: Yeah, I think is only because of the vetkeys library. I can look into it.

marc0olo:
## Review: vetkeys skill

Great direction — replacing the old `vetkd` skill with a focused BLS/IBE/timelock skill and splitting out `encrypted-maps` is the right architecture. A few things to address before merge:

### Description

Remove **"verifiable randomness (VRF)"** from the description. The skill has no VRF content, so an agent routed here for VRF finds nothing actionable. Add it back only if a VRF section is included.

### Cycle costs — precise numbers + missing pitfall

The Key Concepts section uses `~10B` and `~26B`. These should be exact:

- `key_1` mainnet: `26_153_846_153` cycles
- `test_key_1` mainnet: `10_000_000_000` cycles
- Both keys **locally (PocketIC): `26_153_846_153`** cycles — a known deviation; fixing it would require deploying test keys to a dedicated application subnet

The local PocketIC deviation should be a **dedicated pitfall**, not buried in Key Concepts. This is the exact error that caused issue #156. Recommending `30_000_000_000` as a universal safe buffer (unused cycles are refunded) is the simplest advice for agents.

### IBE backend — missing types

"The canister structure is the same as the BLS example" leaves an agent unable to compile. Missing: `SendMessageRequest`, `Message`, `Inbox` struct definitions, the `INBOXES` `StableBTreeMap` declaration, and the `VetKDDeriveKeyArgs` import (used in the snippet but not imported). Either show the full canister or move the IBE example to a `references/` file (like `encrypted-maps` does for Motoko).

### IBE frontend — `identity` undefined

`identity.getPrincipal()` appears with no explanation of where `identity` comes from. The `encrypted-maps` skill handles this correctly — same pattern should apply here.

### Timelock — not compilable as-is

`decrypt_ciphertexts` has `Vec<&[u8]>` in its signature (lifetime parameters can't appear in a Candid endpoint), has no `#[update]` annotation, and calls `close_expired_lots()` which is never defined. This reads as real code but an agent following it will fail to compile. Either provide a complete working example or explicitly mark it as a pseudocode sketch.

### Section ordering

Offline public key derivation sits between timelock and `icp.yaml`, which breaks the reading flow. It belongs inside the IBE section (it's an IBE-specific optimization) or after Deploy as an advanced appendix.

### `icp.yaml` section

The `icp-cli` skill covers recipes, frontend setup, and network config. This section should collapse to the vetKeys-specific part only:

```yaml
canisters:
  - name: backend
    init_args: '("test_key_1")'  # Production: '("key_1")'
```

With a reference: *"For full icp.yaml setup, see the `icp-cli` skill."*

### Deploy & Verify

Drop `icp network start -d` and `icp deploy backend` — those are `icp-cli` territory. Keep the domain-specific verification calls and expected outputs. Add at least one verification step for IBE and timelock, not just BLS.

### Missing: evaluation file

Please add `evaluations/vetkeys.json`. Suggested content:

```json
{
  "skill": "vetkeys",
  "description": "Tests whether agents use the correct vetKeys primitives, avoid raw vetkd_derive_key for BLS, use the correct TypeScript API, and attach sufficient cycles locally.",

  "output_evals": [
    {
      "name": "BLS signing canister",
      "prompt": "I want my canister to sign messages with threshold BLS so clients can verify signatures without trusting the canister. Show me just the Rust backend update function, no deploy steps.",
      "expected_behaviors": [
        "Uses ic_vetkeys::management_canister::sign_with_bls(), not vetkd_derive_key directly",
        "Captures caller before the async call",
        "Returns the signature bytes",
        "Does NOT manually attach cycles — sign_with_bls handles this"
      ]
    },
    {
      "name": "IBE encrypted messaging",
      "prompt": "How do I encrypt a message to a specific user's principal so only they can decrypt it, without exchanging keys first? Show just the frontend encrypt and decrypt flow.",
      "expected_behaviors": [
        "Uses IbeCiphertext.encrypt() with IbeIdentity.fromPrincipal(recipient)",
        "Uses TransportSecretKey.random(), NOT TransportSecretKey.fromSeed()",
        "Uses transportSecretKey.publicKeyBytes(), NOT publicKey()",
        "Uses EncryptedVetKey.deserialize().decryptAndVerify() to get the IBE private key",
        "Does NOT use toDerivedKeyMaterial() or asDerivedKeyMaterial()"
      ]
    },
    {
      "name": "Sealed-bid auction",
      "prompt": "I want bids in my auction canister to stay encrypted until the auction closes, then the canister reveals them. Show the high-level approach and the canister-side decryption logic.",
      "expected_behaviors": [
        "Uses timelock/IBE pattern: identity is the auction or lot ID",
        "Clients encrypt with IbeCiphertext.encrypt() using IbeIdentity.fromBytes(lotId)",
        "Canister decrypts after event using vetkd_derive_key with the event ID as input",
        "Uses a timer (set_timer_interval) to trigger decryption after close",
        "Notes that timers must be re-registered in post_upgrade"
      ]
    },
    {
      "name": "Adversarial: insufficient cycles locally",
      "prompt": "My vetkd_derive_key call fails locally with 'not enough cycles attached' even though I sent 10_000_000_000. What's wrong?",
      "expected_behaviors": [
        "Explains that PocketIC charges 26_153_846_153 cycles for all key names locally",
        "States that test_key_1 only costs 10_000_000_000 on mainnet, not locally",
        "Recommends sending 30_000_000_000 as a safe buffer for both environments",
        "Notes that unused cycles are refunded"
      ]
    }
  ],

  "trigger_evals": {
    "description": "Queries to test whether the vetkeys skill activates correctly.",
    "should_trigger": [
      "I want threshold BLS signatures in my canister",
      "How do I implement identity-based encryption on ICP?",
      "Build a sealed-bid auction on the Internet Computer",
      "I need timelock encryption — data revealed only after an event",
      "How do I encrypt to a recipient without knowing their public key?",
      "Verifiable randomness using vetKeys on ICP"
    ],
    "should_not_trigger": [
      "I want to build a password manager on ICP",
      "How do I add encrypted storage to my canister?",
      "I need user authentication for my dapp",
      "How does stable memory work?",
      "What is EncryptedMaps?"
    ]
  }
}
```

marc0olo:
Superseded by #351, merged as `skills/vetkeys/` (plus `skills/encrypted-maps/` from #158).

@andreacerulli — your scoping of the skill around BLS, IBE and timelock, and the decision to retire the old low-level `vetkd` skill, is what shipped. Thank you; you're credited in #351.

What changed against this branch:

- `@icp-sdk/vetkeys` 0.5 (renamed from `@dfinity/vetkeys`); Rust `ic-vetkeys` 0.9 + `ic-cdk` 0.20 + `ic-cdk-management-canister`; Motoko `ic-vetkeys` 0.6 — which answers the inline question here about the core library: the Motoko library now exists, so the management canister is no longer hand-rolled
- The feature depth that was inline (and incomplete) moved to `references/ibe.md` (IBE + timelock, with the full type definitions) and `references/bls-signing.md` — the two gaps flagged in the review
- Exact cycle costs, and `test_key_1` / `key_1` behave the same locally and on mainnet
- VRF is now an actual section rather than only a description claim
- `evaluations/vetkeys.json` shipped with it (5 output cases + bidirectional trigger routing against `encrypted-maps`)

Closing as superseded.

PR #170, opened 2026-04-16
fix: Add warning about the caching of skills

raymondk:


marc0olo:
The new `## Important` section conflicts with step 2, which currently says:

> "use it if already loaded in your context"

An agent reading both instructions gets contradictory guidance — one says reuse what's in context, the other says don't use cached versions blindly.

Rather than adding a separate section, I'd suggest folding the intent directly into step 2 and dropping the phrase that causes the conflict:

**Current:**
```
2. When a task matches a skill's description, use it if already loaded in your
   context. Otherwise, fetch the skill content from its url.
   Do not construct skill URLs manually.
```

**Suggested:**
```
2. When a task matches a skill's description, fetch the skill content from its url
   before using it. Skills are updated frequently — do not rely on a previously
   cached or trained version.
   Do not construct skill URLs manually.
```

Stale skill content causing broken code is a worse tradeoff than an extra HTTP request, so dropping "use it if already loaded" is the right call.

marc0olo (approved):
this now removes the following sentence:
> Do not construct skill URLs manually.

but I don't think it is an issue anymore since we include the URL now explicitly and even provide an example for the II skill.

PR #171, opened 2026-04-16
refactor: restructure skill markdown URLs and improve the site

raymondk:
- return markdown urls from .well-known
- add note about not relying on cache
- fix the text on how it works
- add the source commit for a skill
- add times and commit to skill


marc0olo:
Looking good. Two remaining things:

- `api/skills.json.ts` and `api/skills/[slug].json.ts` still use `githubUrl()` (main branch) for `urls.source`. Now that the skill page shows a commit-pinned link, the JSON API is out of sync — `updated` date and `source` URL describe different versions. Worth switching to `githubCommitUrl` (needs `getSkillGitInfo` in those two files).

- Nit: the new `.ic-assets.json5` entry is commented `"User discovery endpoints"` — the adjacent rule says `"Agent discovery endpoints"`. Suggest something like `"Raw skill markdown endpoints"` to match what it actually is.

marc0olo:
Thanks — `api/skills/[slug].json.ts` looks good now. `api/skills.json.ts` (the full index) still uses `githubUrl()` for `urls.source`. Since `getSkillUpdatedAt` already calls `getSkillGitInfo` internally, the SHA is already being computed and discarded — a small refactor to destructure it would make the index consistent too. Low priority, but worth tracking.

marc0olo (approved):

PR #172, opened 2026-04-16
fix: Remove deprecated functions and return utc time

raymondk:


marc0olo:
Clean follow-up, addresses all the feedback. One nit: `getSkillCommitHash` (added in #171) appears to be dead code — nothing calls it, everywhere uses `getSkillGitInfo` with destructuring directly. Safe to remove.

marc0olo (approved):

PR #176, opened 2026-04-28
chore: Add information about using mainnet identities with pocket ic.

raymondk:


marc0olo (approved):
I think we can improve the text a bit for targeting agents specifically.

But I don't think it matters too much.

PR #198, opened 2026-05-29
feat: integrate skill-creator, add improve-ic-skill, overhaul contributor workflow

marc0olo:
### New: skill-creator installed
`skill-creator` vendored at `.agents/skills/skill-creator/` with a symlink at `.claude/skills/skill-creator` for Claude Code auto-discovery. Installed via `npx skills add` from `anthropics/skills` commit `b0cbd3df` (2026-03-06).

`skills-lock.json` is gitignored — our copy is a patched fork and must not be silently overwritten by `npx skills add`. Re-apply patches from `PATCHES.md` if updating intentionally.

**5 bugs patched** (confirmed from testrun, documented in `.agents/skills/skill-creator/PATCHES.md`):
1. `generate_review.py`: escape `</script>` — viewer breaks on HTML eval output containing script tags
2. `SKILL.md`: add missing `run-1/` path level — `aggregate_benchmark.py` silently drops all results without it
3. `SKILL.md`: require `eval-0-descriptive-name/` format — aggregator globs `eval-*`, purely descriptive names vanish
4. `SKILL.md`: `--static` wrong for Claude Code — generates `file://` URLs browsers restrict; use server mode
5. `quick_validate.py`: replace PyYAML (undeclared third-party dep) with stdlib parser — `package_skill.py` fails at import in a clean environment

**1 doc improvement** to `agents/grader.md`: explicit warning that `grading.json` expectations must use `text`/`passed`/`evidence` field names — viewer silently shows empty grades otherwise.

### New: improve-ic-skill
Internal skill at `.agents/skills/improve-ic-skill/` for improving existing skills. Token-efficient alternative to skill-creator's heavy interactive loop.

Key behaviours:
- Step 0 hard gate: requires an explicit problem statement before doing any work
- Knows our toolchain (`npm run validate`, `evaluate-skills.js` flags), eval location (`evaluations/<skill-name>.json`), and upstream-tracked skill rules
- Distinguishes updating the eval file (Step 6) from running existing evals (Step 7) — evals are run selectively, not on every pass
- Seeds evals from the upstream diff when a skill has none yet
- Routes upstream sync work through this skill

### Deleted: legacy skill template
`skills/_template/SKILL.md.template` removed — its prescribed body sections contradicted the "no rigid structure" principle and skill-creator now handles drafting.

### Updated: `.gitignore`
Added entries for skill-creator runtime artifacts: `skills/*-workspace/`, `**/__pycache__/`, `skills-lock.json`.

### Updated: `CLAUDE.md` and `CONTRIBUTING.md`
- `improve-ic-skill` for improving existing skills; `skill-creator` for creating new skills — explicit in both docs
- Two-phase eval workflow clarified: skill-creator's internal loop for drafting; `evaluations/<skill-name>.json` + `evaluate-skills.js` as the committed regression safety net
- Eval run requirements: required with baseline for new skill PRs; recommended-not-required for improvements
- Upstream sync: `improve-ic-skill` named as the skill to load; eval guidance aligned with general policy; branch naming clarified (`<repo>` = upstream repo short name)
- Project Structure corrected (`src/data/` → `src/lib/`, non-existent components removed)
- ~15 additional consistency fixes (step numbering, diff notation, eval path references, category instructions, PR eval scope, branch naming)

### Not fixed (intentional)
`grader.md` step ordering bug (Step 7 writes before Step 8 reads metrics) — pre-existing in upstream Anthropic code, does not affect `improve-ic-skill`'s workflow.


marc0olo:
needs to be tested properly before merging

.agents/skills/skill-creator/eval-viewer/generate_review.py
```diff
@@ -0,0 +1,471 @@
...
+        "runs": runs,
+        "previous_feedback": previous_feedback,
+        "previous_outputs": previous_outputs,
+    }
+    if benchmark:
+        embedded["benchmark"] = benchmark
+
+    data_json = json.dumps(embedded)
+
+    return template.replace("/*__EMBEDDED_DATA__*/", f"const EMBEDDED_DATA = {data_json};")
```
marc0olo, line 283: **Bug: `</script>` in eval output breaks the viewer entirely**

`json.dumps` does not escape `<` or `/`, so if any eval output contains an HTML `<script>` tag (e.g. an `index.html` example with `<script type="module" ...></script>`), the literal `</script>` closes the `<script>` block in `viewer.html` prematurely and everything after it renders as raw text — the entire viewer breaks.

Confirmed in practice: the full-stack hello-world eval output includes `index.html` with a `</script>` tag, which broke the viewer completely.

**Fix:**
```python
data_json = json.dumps(embedded)
data_json = data_json.replace("</script>", r"<\/script>")
return template.replace("/*__EMBEDDED_DATA__*/", f"const EMBEDDED_DATA = {data_json};")
```

.agents/skills/skill-creator/SKILL.md
```diff
@@ -0,0 +1,485 @@
...
+- Input files: <eval files if any, or "none">
+- Save outputs to: <workspace>/iteration-<N>/eval-<ID>/with_skill/outputs/
+- Outputs to save: <what the user cares about — e.g., "the .docx file", "the final CSV">
+```
+
+**Baseline run** (same prompt, but the baseline depends on context):
+- **Creating a new skill**: no skill at all. Same prompt, no skill path, save to `without_skill/outputs/`.
+- **Improving an existing skill**: the old version. Before editing, snapshot the skill (`cp -r <skill-path> <workspace>/skill-snapshot/`), then point the baseline subagent at the snapshot. Save to `old_skill/outputs/`.
+
+Write an `eval_metadata.json` for each test case (assertions can be empty for now). Give each eval a descriptive name based on what it's testing — not just "eval-0". Use this name for the directory too. If this iteration uses new or modified eval prompts, create these files for each new eval directory — don't assume they carry over from previous iterations.
```
marc0olo: **Naming conflict: `eval-0/` vs descriptive directory name**

Line 167 says directories should be `eval-0/`, `eval-1/`, etc. Line 188 contradicts this:

> Give each eval a descriptive name based on what it's testing — not just `eval-0`. Use this name for the directory too.

Following line 188 literally produces directories like `full-stack-motoko-hello-world/` which `aggregate_benchmark.py` does not find because it globs for `eval-*`.

The intended convention is `eval-0-descriptive-name/` (numeric prefix required by the script, descriptive suffix for humans). Neither line says this.

**Fix:** Change line 188 to: "Give each eval a descriptive name. Always prefix with the numeric ID so the tools can find it: `eval-0-full-stack-setup/`, `eval-1-migration/`. Use this full prefixed name for the directory."

.agents/skills/skill-creator/SKILL.md
```diff
@@ -0,0 +1,485 @@
...
+For each test case, spawn two subagents in the same turn — one with the skill, one without. This is important: don't spawn the with-skill runs first and then come back for baselines later. Launch everything at once so it all finishes around the same time.
+
+**With-skill run:**
+
+```
+Execute this task:
+- Skill path: <path-to-skill>
+- Task: <eval prompt>
+- Input files: <eval files if any, or "none">
+- Save outputs to: <workspace>/iteration-<N>/eval-<ID>/with_skill/outputs/
```
marc0olo: **Missing `run-N/` level in the output path, causing `(No prompt found)` in viewer**

The path shown here is `eval-<ID>/with_skill/outputs/`. The actual required structure is `eval-<ID>/with_skill/run-1/outputs/`.

Two concrete consequences:
1. `aggregate_benchmark.py` globs for `run-*` inside config directories — it skips any config dir that has no `run-N/` subdirectory, so all runs show 0% pass rate.
2. The viewer looks for `eval_metadata.json` at `run_dir.parent` — which is `with_skill/` when using `run-1/`. If the path is `with_skill/outputs/` (no `run-1/`), the viewer treats `with_skill` itself as the run dir and looks for metadata at `with_skill.parent = eval-0/` — two hops up, and only works if the file is there. But if the path is `with_skill/run-1/outputs/`, the viewer looks at `run-1.parent = with_skill/`, which means `eval_metadata.json` must be copied to each config directory, not just the eval directory root.

**Fix:** Change the path to `eval-<ID>/with_skill/run-1/outputs/` and add a note: "Also copy `eval_metadata.json` into each config directory (`with_skill/`, `old_skill/`) so the viewer can display the eval prompt."

.agents/skills/skill-creator/SKILL.md
```diff
@@ -0,0 +1,485 @@
...
+   nohup python <skill-creator-path>/eval-viewer/generate_review.py \
+     <workspace>/iteration-N \
+     --skill-name "my-skill" \
+     --benchmark <workspace>/iteration-N/benchmark.json \
+     > /dev/null 2>&1 &
+   VIEWER_PID=$!
+   ```
+   For iteration 2+, also pass `--previous-workspace <workspace>/iteration-<N-1>`.
+
+   **Cowork / headless environments:** If `webbrowser.open()` is not available or the environment has no display, use `--static <output_path>` to write a standalone HTML file instead of starting a server. Feedback will be downloaded as a `feedback.json` file when the user clicks "Submit All Reviews". After download, copy `feedback.json` into the workspace directory for the next iteration to pick up.
```
marc0olo: **`--static` should not be used in Claude Code**

This line says to use `--static` when `webbrowser.open()` is not available or "the environment has no display." In Claude Code (VS Code extension, desktop app), the environment has a display and `webbrowser.open()` works — but an agent may not know that, and defaults to the conservative `--static` path.

`--static` generates a file opened via `file://` URL. Browsers block certain JavaScript APIs for `file://` pages, causing the viewer to fail with the same symptom as the `</script>` bug: the UI renders but outputs are not populated.

**Fix:** Clarify when each mode applies:

> In Claude Code (VS Code extension or desktop app), use server mode — do **not** use `--static`. Only use `--static` in true headless environments (Cowork, CI, remote server with no graphical display). If you are unsure, try server mode first."

marc0olo:
**Two additional items not tied to specific lines:**

**1. Missing `.gitignore` entries for skill-creator artifacts**

Running skill-creator in this repo produces two untracked directories that are not currently gitignored:
- `skills/*-workspace/` — the iteration workspace (CLAUDE.md already says it's not committed, but nothing prevents it from being accidentally staged)
- `**/__pycache__/` — Python bytecode generated when running the aggregation and viewer scripts

Suggest adding both to `.gitignore` as part of this PR (or a follow-up).

---

**2. `agents/grader.md` doesn't mention the field name requirement for `grading.json`**

SKILL.md line 225 warns that grading.json expectations must use `text`, `passed`, `evidence` (not `name`/`met`/`details`). But a grader subagent reads only `agents/grader.md` — and that file's output format section doesn't repeat this constraint. If a grader subagent uses different field names, the viewer silently shows empty grades with no error.

The warning should appear in `agents/grader.md` directly above the output format JSON example, not just in SKILL.md.

marc0olo:
here is what happens if I ask the agent to improve the icp-cli skill without giving clear guidance:
image

marc0olo:
after we fixed the generic "just improve it" prompt:

image

marc0olo:
seems like we are good now on this:
image

raymondk (approved):

PR #226, opened 2026-06-17
feat: adopt icp.net as the mainnet HTTP gateway domain (icp-cli v1.0.0)

### LINKED ISSUE #224: Adopt icp.net as the mainnet HTTP gateway domain (icp-cli v1.0.0)
## Context

icp-cli **v1.0.0** changed the default mainnet HTTP gateway domain from `icp0.io` to `icp.net` (`IC_MAINNET_NETWORK_GATEWAY_URL = "https://icp.net"`). The browser canister-access domain is now `https://<canister-id>.icp.net` (and `raw.icp.net`).

The developer-docs have been updated to use `icp.net` for all browser/HTTP-gateway canister-access URLs (verified: `<id>.icp.net`, `<id>.raw.icp.net`, and the `icp.net/custom-domains/v1/` API all resolve and serve correctly). The skills here still reference `icp0.io` / `ic0.app` and should be reconciled so docs and skills stay consistent.

## Skill files referencing the old domains

- `skills/asset-canister/SKILL.md` — certified vs raw serving framed as `ic0.app` / `icp0.io` and `raw.ic0.app` / `raw.icp0.io`
- `skills/custom-domains/SKILL.md`
- `skills/icp-cli/SKILL.md`
- `skills/internet-identity/SKILL.md` — see caveat below
- `skills/certified-variables/SKILL.md`
- `skills/deploy-to-cloud-engine/SKILL.md`
- `skills/agent-web-identity/SKILL.md`

## What to change vs. keep

**Change → `icp.net`:** browser/HTTP-gateway canister-access URLs (`<id>.icp0.io`/`<id>.ic0.app`), `raw.*` domains, and illustrative frontend origins.

**Keep (NOT the gateway):**
- **Agent / API hosts** — `HttpAgent` host, `create_agent`, `.with_url`, anything hitting `/api/v2`. These use the **API** endpoint (`icp-api.io`), which is separate from the gateway. `icp.net` is gateway-only.
- **Named dapp domains** — `nns.ic0.app`, `identity.ic0.app`.
- **II domain-rewriting behavior** — the II skill currently states "II automatically rewrites `icp0.io` to `ic0.app` during delegation." Confirm how `icp.net` participates in this rewriting before changing the behavioral claim; it should not become a meaningless "icp.net vs icp.net" statement.

## Cross-reference

Mirrors the change already applied in dfinity/developer-docs (gateway domain + icp-cli v1.0.0 / Motoko recipe v5.0.0 bump).

marc0olo:
## Summary

Resolves #224. icp-cli **v1.0.0** changed the default mainnet HTTP gateway from `icp0.io` to `icp.net` (browser canister access is now `https://<canister-id>.icp.net` and `raw.icp.net`). This reconciles the skills with the [developer-docs change](https://github.com/dfinity/developer-docs) (mirror PR open there).

## Changed → `icp.net` (gateway / browser canister-access URLs)

| File | What changed |
|------|--------------|
| `asset-canister` | mainnet access-URL table, certified/raw serving domains, browser-open URL; **plus** the Node.js `HttpAgent` host changed from the legacy `ic0.app` access domain to the canonical API endpoint `icp-api.io` (the rest of the corpus already uses `icp-api.io` as the agent host) |
| `custom-domains` | default access URL, `.well-known/ic-domains` fetch URLs, and the `custom-domains/v1` REST API host (issue confirms it resolves on `icp.net`) |
| `certified-variables` | mainnet `curl` example |
| `deploy-to-cloud-engine` | frontend canister access URL |
| `internet-identity` | illustrative `frontend_origins` / `TRUSTED_ORIGIN` examples |

## `custom-domains` HttpAgent `host` guidance

Pitfall 8 and the **HttpAgent Configuration** section are corrected: a recent `@icp-sdk/core` `HttpAgent` resolves an omitted `host` to `https://icp-api.io` on a custom domain, so you do **not** need to set it. The guidance now warns against the real mistake — pointing `host` at your custom domain / `window.location.origin`, which is the gateway and does not serve `/api/v2`. Scoped to the mainnet custom-domains service only.

## Kept (NOT the gateway)

- **API / agent hosts** stay `icp-api.io` — `icp.net` is gateway-only.
- **Named dapp domains** (`nns.ic0.app`) in `icp-cli` and `agent-web-identity`.
- **II delegation rewriting pitfall** (`internet-identity`): II still rewrites `icp0.io` → `ic0.app` at the protocol level (per the II spec — unchanged by the gateway rename), so this behavioral claim is left as-is rather than turned into a meaningless "icp.net vs icp.net" statement. This is the caveat the issue flagged.
- `icp-cli` and `agent-web-identity` SKILL.md needed **no** changes (only API host + named dapp domains).

## Evals

- Updated `custom-domains` API-endpoint assertions and the `asset-canister` mainnet-host behavior to match the domain change.
- Updated the `custom-domains` "HttpAgent host configuration" case to match the corrected `host` guidance.
- Added an `asset-canister` regression guard: the mainnet browser URL must be `<id>.icp.net` (not `ic0.app`/`icp0.io`/`icp-api.io`).
- The `internet-identity` rewriting eval is intentionally left referencing `icp0.io`/`ic0.app` since the pitfall it tests is unchanged.

<details>
<summary>Eval results (new/changed cases)</summary>

```
asset-canister — "Mainnet browser-access domain is icp.net": WITH 3/3
  ✅ Gives the URL as https://<canister-id>.icp.net
  ✅ Does NOT use the legacy ic0.app or icp0.io gateway domains
  ✅ Does NOT use icp-api.io (that is the API endpoint, not the browser gateway)

custom-domains — "Domain registration API": WITH 4/4
  ✅ Uses the validate endpoint: GET https://icp.net/custom-domains/v1/app.example.com/validate
  ✅ Uses the registration endpoint: POST https://icp.net/custom-domains/v1/app.example.com
  ✅ Uses the status check endpoint: GET https://icp.net/custom-domains/v1/app.example.com
  ✅ Does NOT invent non-existent API endpoints or parameters

custom-domains — "HttpAgent host configuration": WITH 3/3
  ✅ Identifies that the custom domain serves only the HTTP gateway, not the /api/v2 API endpoint, so host must not point at it
  ✅ Says to leave host unset (defaults to https://icp-api.io on a custom domain) or set it explicitly to https://icp-api.io
  ✅ Does NOT recommend setting host to the custom domain or window.location.origin
```
</details>

`npm run validate` → all 26 skills pass.

Closes #224

### DIFF
--- a/skills/custom-domains/SKILL.md
+++ b/skills/custom-domains/SKILL.md
@@ -40,7 +40,7 @@ Custom domains work at the boundary node level — they map a domain to any cani
 
 7. **Not explicitly registering the domain.** DNS configuration alone is not enough. You must call `POST /custom-domains/v1/CUSTOM_DOMAIN` to start registration. It is not automatic.
 
-8. **Not setting `host` in HttpAgent on custom domains.** When serving from a custom domain, the `HttpAgent` cannot automatically infer the IC API host like it can on `icp.net`. You must set `host: "https://icp-api.io"` explicitly for mainnet.
+8. **Pointing `HttpAgent`'s `host` at your custom domain.** You usually do not need to set `host` at all. When `host` is omitted, a recent `HttpAgent` (`@icp-sdk/core`) defaults to the local replica for `localhost`/`127.0.0.1` origins and to `https://icp-api.io` (the mainnet API boundary nodes) for every other origin — including custom domains and `icp.net`. The actual mistake is forcing `host` to your custom domain (or `window.location.origin`): that domain serves only the HTTP gateway, not `/api/v2`, so canister calls fail. Leave `host` unset, or set it explicitly to `https://icp-api.io` — never the gateway domain.
 
 9. **Forgetting alternative origins for Internet Identity.** II principals depend on the origin domain. Switching from a canister URL to a custom domain changes principals. Configure `.well-known/ii-alternative-origins` to keep the same principals. See the `internet-identity` skill.
 
@@ -178,14 +178,19 @@ curl -sL -X GET "https://icp.net/custom-domains/v1/CUSTOM_DOMAIN" | jq
 
 ## HttpAgent Configuration
 
-On custom domains, the agent cannot auto-detect the IC API host. Set it explicitly:
+You usually do not need to set `host`. When `host` is omitted, a recent `HttpAgent` defaults to the local replica for `localhost`/`127.0.0.1` origins and to `https://icp-api.io` (the mainnet API boundary nodes) for every other origin — including custom domains and `icp.net`. Do **not** point `host` at your custom domain: it serves only the HTTP gateway, not `/api/v2`, so calls would fail.
 
 ```typescript
 import { HttpAgent } from "@icp-sdk/core/agent";
 
-const isProduction = process.env.NODE_ENV === "production";
-const host = isProduction ? "https://icp-api.io" : undefined;
-const agent = await HttpAgent.create({ host });
+// host omitted → local replica on localhost origins, https://icp-api.io elsewhere
+const agent = await HttpAgent.create();
+```
+
+In Node.js there is no `window.location` to detect the origin, so an omitted `host` always resolves to `https://icp-api.io`. Set `host` to your local replica URL when talking to a local network; for mainnet you can leave it unset or set it explicitly to the API endpoint (never the gateway):
+
+```typescript
+const agent = await HttpAgent.create({ host: "https://icp-api.io" });
 ```
 
 ## Deploy & Test
--- a/skills/custom-domains/SKILL.md
+++ b/skills/custom-domains/SKILL.md
@@ -40,7 +40,7 @@ Custom domains work at the boundary node level — they map a domain to any cani
 
 7. **Not explicitly registering the domain.** DNS configuration alone is not enough. You must call `POST /custom-domains/v1/CUSTOM_DOMAIN` to start registration. It is not automatic.
 
-8. **Pointing `HttpAgent`'s `host` at your custom domain.** You usually do not need to set `host` at all. When `host` is omitted, a recent `HttpAgent` (`@icp-sdk/core`) defaults to the local replica for `localhost`/`127.0.0.1` origins and to `https://icp-api.io` (the mainnet API boundary nodes) for every other origin — including custom domains and `icp.net`. The actual mistake is forcing `host` to your custom domain (or `window.location.origin`): that domain serves only the HTTP gateway, not `/api/v2`, so canister calls fail. Leave `host` unset, or set it explicitly to `https://icp-api.io` — never the gateway domain.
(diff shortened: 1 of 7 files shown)


skills/custom-domains/SKILL.md
```diff
@@ -40,7 +40,7 @@ Custom domains work at the boundary node level — they map a domain to any cani
 
 7. **Not explicitly registering the domain.** DNS configuration alone is not enough. You must call `POST /custom-domains/v1/CUSTOM_DOMAIN` to start registration. It is not automatic.
 
-8. **Not setting `host` in HttpAgent on custom domains.** When serving from a custom domain, the `HttpAgent` cannot automatically infer the IC API host like it can on `icp0.io`. You must set `host: "https://icp-api.io"` explicitly for mainnet.
+8. **Not setting `host` in HttpAgent on custom domains.** When serving from a custom domain, the `HttpAgent` cannot automatically infer the IC API host like it can on `icp.net`. You must set `host: "https://icp-api.io"` explicitly for mainnet.
```
raymondk: I don't think this is accurate -https://github.com/dfinity/icp-js-core/blob/3d2c1079f0ced355a623b54bc19be566d6431755/packages/core/src/agent/agent/http/index.ts#L215

It will default to icp-api.io you don't need to set that.

marc0olo: Good catch — you're right, and the original premise was inaccurate (it predates this PR; my change just touched the same line). I confirmed against `determineHost()` in `@icp-sdk/core`: with `host` omitted it defaults to the local replica for `localhost`/`127.0.0.1` origins and to `https://icp-api.io` for every other origin — including custom domains and `icp.net` (neither is in `knownHosts`). So on a custom domain you don't need to set `host`; it already resolves to the API boundary nodes.

I've reworked pitfall 8 and the **HttpAgent Configuration** section accordingly: the guidance now says you usually don't set `host`, explains the default behavior, and reframes the real mistake as pointing `host` at your custom domain / `window.location.origin` (which serves only the gateway, not `/api/v2`). `icp-api.io` stays as the explicit option (and the only thing to set in Node.js local dev). Updated the matching eval case too (now 3/3). See 5e6c01d.

marc0olo: Expanded the section to make every case unambiguous (164b959). The **HttpAgent Configuration** section now leads with "`host` is the API endpoint, not the domain the frontend is served from", then gives a resolution table for what an omitted `host` resolves to per runtime, and a per-scenario "what to do":

- **Mainnet browser frontend** (`icp.net` / custom domain / legacy `ic0.app`·`icp0.io`): leave `host` unset → resolves to `icp-api.io`. Never set it to the gateway/custom domain.
- **Mainnet Node.js**: unset or `icp-api.io`.
- **Local dev, browser on `*.localhost`**: unset → auto-detects the local replica.
- **Local dev, Node.js**: *must* set `host` to the local replica URL — no `window.location`, so omitted `host` defaults to `icp-api.io` (mainnet).
- **Non-mainnet / custom network**: *must* set `host` to that network's API endpoint — the `icp-api.io` default points at mainnet.

That last one is the case @marc0olo raised — a custom *network* (not the mainnet custom-domains service) genuinely needs an explicit `host`, so it's now documented. Added an eval for the Node.js-local trap too. All HttpAgent eval cases pass 3/3.

marc0olo: Trimmed back to mainnet-only (3bd20f0). The previous revision added local-dev and custom-network cases, but those are out of scope for this skill — the custom-domains service is mainnet-only (you can't have a custom domain on a local replica or a custom network), so that material just diluted it.

The **HttpAgent Configuration** section now says only what's relevant here: a mainnet frontend served from a custom domain doesn't need `host` set (it resolves to `https://icp-api.io`), and must never point `host` at the gateway/custom domain. Pitfall 8 matches. Dropped the out-of-scope Node.js eval; the custom-domain eval still passes 3/3.

raymondk (approved):

PR #244, opened 2026-07-13
docs(deploy-to-cloud-engine): triage sandboxed agent environments before identity linking

artkorotkikh-dfinity:
# Motivation

Agents running the skill from a sandboxed environment hit a wall at Step 1: `icp identity link web` completes the sign-in via a redirect to `127.0.0.1:<port>` on the CLI host, so when the CLI runs in a cloud sandbox (seen in the wild with Perplexity) the user's browser can never deliver the delegation, and the agent keeps retrying a flow that cannot succeed. A second variant is a device-bridge shell on the user's own machine with no network at all, where neither the link nor the deploy can run.

# Changes

- Added Step 1.0 environment triage: normal flow, remote sandbox with network, and no-network shell, each with the correct path.
- Added the delegation handoff for remote sandboxes (`icp identity delegation request` / `sign` / `use`), verified end-to-end on icp-cli 1.0.2 with matching principals, including the `reauth` recovery when the local web session has expired.
- Added a one-script handoff for no-network bridge shells: the agent prepares `deploy.sh`, the user runs it in their real terminal where the normal link flow works.
- Added pitfalls 14 (browser and CLI don't share localhost) and 15 (delegation handoff doesn't fix a missing network).
- Added a macOS TCC note to Prerequisites (`~/Library/Application Support/org.dfinity.icp-cli/`, `HOME` override caveat) and bumped verified versions to 0.3.0 and 1.0.2.
- Added an install hint to Related Skills so agents outside the icskills ecosystem can resolve cross-referenced skills.
- Updated the description with sandbox-failure trigger keywords (1021/1024 chars).
- Added two adversarial eval cases and a should-trigger query covering both sandbox variants.

# Eval results

<details>
<summary>Output eval 6 — Adversarial: CLI in a remote sandbox, browser on another machine (WITH 5/5 | WITHOUT 2/5)</summary>

```
  WITH skill: 5/5 passed
    ✅ Identifies that the callback targets the sandbox's loopback, which the laptop browser cannot reach
    ✅ Does NOT suggest re-running `icp identity link web` in the sandbox or opening the 127.0.0.1 URL again
    ✅ Recommends the delegation handoff flow (request/sign/use with --key-pem, --duration, --from-json)
    ✅ Does NOT invent a headless/no-browser flag on `icp identity link web`
    ✅ Treats the signed delegation as a time-limited credential (bounded --duration)

  WITHOUT skill: 2/5 passed
    ✅ Identifies that the callback targets the sandbox's loopback, which the laptop browser cannot reach
    ❌ Does NOT suggest re-running `icp identity link web` in the sandbox or opening the 127.0.0.1 URL again
       → Option A explicitly instructs the user to re-run `icp identity link web admin --auth https://opencloud.org` in the sandbox and complete sign-in again via a tunnel.
    ❌ Recommends the delegation handoff
       → The output never mentions the delegation request/sign/use subcommands; it proposes SSH tunneling or manually copying identity files instead.
    ✅ Does NOT invent a headless/no-browser flag on `icp identity link web`
    ❌ Treats the signed delegation as a time-limited credential (bounded --duration)
       → Since the delegation flow is never mentioned, there is no discussion of a bounded --duration or time-limited credential.
```

</details>

<details>
<summary>Output eval 7 — Adversarial: bridge shell on the user's machine has no network (WITH 5/5 | WITHOUT 3/5)</summary>

```
  WITH skill: 5/5 passed
    ✅ Identifies that the delegation handoff cannot help here: it moves signing authority, not network access
    ✅ Recommends preparing a single script the user runs in their real Terminal with icp identity link web, icp identity default, icp deploy -e ic --subnet
    ✅ Asks the user to paste the script output back for verification
    ✅ Does NOT suggest tunneling, proxying, or retrying the commands inside the sandboxed shell
    ✅ Does NOT use dfx

  WITHOUT skill: 3/5 passed
    ❌ Identifies that the delegation handoff cannot help here: it moves signing authority, not network access
       → The output correctly notes both commands need network and the sandbox lacks it, but never addresses the specific conceptual point that delegation transfers signing authority rather than network access.
    ✅ Recommends preparing a single script the user runs in their real Terminal, doing the normal flow there
    ❌ Asks the user to paste the script output back for verification
       → The output ends with a caveat about the --auth origin but never asks the user to report back or paste command output for verification.
    ✅ Does NOT suggest tunneling, proxying, or retrying the commands inside the sandboxed shell
    ✅ Does NOT use dfx
```

</details>

<details>
<summary>Trigger evals — 11/11 should-trigger, 8/8 should-not-trigger (rewritten description)</summary>

```
  Should trigger: 11/11 correct
    ✅ "How do I deploy this app to my cloud engine?"
    ✅ "Push my canisters to my OpenCloud engine"
    ✅ "Link the icp CLI to my engine console"
    ✅ "Deploy to my engine on a specific subnet"
    ✅ "How do I authorize the icp CLI for my cloud engine?"
    ✅ "Ship this project to my cloud engine"
    ✅ "My agent should deploy straight to my engine"
    ✅ "My agent runs in a cloud sandbox and icp identity link web never completes — the sign-in redirects to 127.0.0.1"
    ✅ "Make my deployed app show up with a name in the cloud engine console"
    ✅ "Label my backend and frontend canisters in the engine console instead of bare principal ids"
    ✅ "Give my deployed app an icon in the OpenCloud engine console"

  Should NOT trigger: 8/8 correct
    ✅ "Package my app as a .icp archive so others can install it on their engines from the marketplace"
    ✅ "Deploy my canister to mainnet"
    ✅ "Set up icp.yaml for my Rust canister"
    ✅ "Add Internet Identity login to my frontend"
    ✅ "How does stable memory work in Rust canisters?"
    ✅ "Implement an ICRC-1 token transfer"
    ✅ "Generate TypeScript bindings for my canister"
    ✅ "Add access control to my Motoko canister"
```

</details>

### DIFF
--- a/evaluations/deploy-to-cloud-engine.json
+++ b/evaluations/deploy-to-cloud-engine.json
@@ -55,6 +55,28 @@
       ]
     },
     {
+      "name": "Adversarial: CLI in a remote sandbox, browser on another machine",
+      "prompt": "My coding agent runs in a cloud sandbox. It ran `icp identity link web admin --auth https://opencloud.org` there and gave me a sign-in URL containing `callback=http://127.0.0.1:39281/`. I completed the Internet Identity sign-in in the browser on my laptop, but the CLI in the sandbox never finishes linking. What is wrong and how do I get the sandbox CLI authorized for my engine? Just the fix, no deploy steps.",
+      "expected_behaviors": [
+        "Identifies that the callback targets the sandbox's loopback, which the laptop browser cannot reach — the browser resolves 127.0.0.1 to the laptop itself, so the delegation never reaches the CLI",
+        "Does NOT suggest re-running `icp identity link web` in the sandbox or opening the 127.0.0.1 URL again — that can never complete",
+        "Recommends the delegation handoff: `icp identity delegation request <name>` in the sandbox (prints a session public key PEM), `icp identity delegation sign --key-pem <file> --duration <d>` on the laptop with a locally web-linked identity, then `icp identity delegation use <name> --from-json <file>` back in the sandbox (or running the whole flow on the laptop instead)",
+        "Does NOT invent a headless/no-browser flag on `icp identity link web`",
+        "Treats the signed delegation as a time-limited credential (bounded --duration)"
+      ]
+    },
+    {
+      "name": "Adversarial: bridge shell on the user's machine has no network",
+      "prompt": "My coding agent has shell access on my Mac through a sandboxed device bridge: DNS is blocked (curl returns status 000) and it can only write to /tmp. It already built my ICP project. It proposed doing an identity delegation handoff and then running the deploy to my cloud engine itself from that shell. Is that going to work, and if not, what should it do instead? I know my subnet id. Just the approach and commands.",
+      "expected_behaviors": [
+        "Identifies that the delegation handoff cannot help here: it moves signing authority, not network access — `icp deploy` (and the link) still need network from the shell that runs them, so nothing network-bound can run in the bridge shell",
+        "Recommends preparing a single script the user runs in their real Terminal, doing the normal flow there: `icp identity link web <name> --auth <console-origin>` (terminal and browser share a loopback, so no handoff is needed), `icp identity default <name>`, `icp deploy -e ic --subnet <subnet-id>`",
+        "Asks the user to paste the script output back for verification",
+        "Does NOT suggest tunneling, proxying, or retrying the commands inside the sandboxed shell",
+        "Does NOT use dfx"
+      ]
+    },
+    {
       "name": "Agent pitfall: link command stalls non-interactively",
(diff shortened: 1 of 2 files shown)


marc0olo (approved):
Approve — clean, well-structured sandbox-triage addition. Verified: description is 1021/1024 chars, eval JSON parses (14 output evals + triggers intact), delegation identity-name placeholders and flags (`--key-pem`/`--duration`/`--from-json`/`--storage`) are internally consistent with the eval expectations, Pitfall 1→14 cross-reference resolves, and the `set -euo pipefail` + `grep … || icp identity link` script logic is correct.

Non-blocking note: the description's "Use when" clause drops the explicit proxy-canister trigger (to fit under 1024 chars), and the trigger-eval set has no proxy/XRC/threshold-signing query — so proxy triggering is unmeasured. Risk is low since the proxy keywords survive in the "Also covers …" sentence, and this coverage gap predates this PR, so no need to address it here. Worth a small follow-up should-trigger case next time these evals are touched.

PR #257, opened 2026-07-30
autosync: atomic skill swap so intra-skill file renames/removals don't orphan

marc0olo:
## What & why

The autosync sync script (`.claude/sync-ic-skills.sh`) is already robust at the **skill level** — every session it re-fetches the full discovery index, downloads newly-published skills, and prunes managed skills that vanished from the index. So a whole-skill add/rename/delete is handled automatically, and it avoids the gaps that affect the version-pinned `npx skills` flow entirely:

- [vercel-labs/skills#591](https://github.com/vercel-labs/skills/issues/591) — `skills update` never discovers newly-published skills.
- [vercel-labs/skills#1376](https://github.com/vercel-labs/skills/issues/1376) — the stale-skill prune silently no-ops for shorthand sources (`owner/repo`).

But there was one gap left, **one level down**: when a skill's `hash` changed, the script re-downloaded its current files *over the existing directory without clearing it*. So if a new version of a skill **renamed or removed a file** (e.g. dropped a `references/*.md`), the stale orphan lingered inside the skill dir.

## Fix

Download each changed skill into a staging dir, then swap it in **atomically on full success** (mirroring the script's existing `mv "$NEW_MANIFEST" "$MANIFEST"` pattern):

- Files removed/renamed upstream no longer survive locally — the skill dir is replaced wholesale.
- A mid-download failure keeps the **existing** copy intact (the swap happens only after every file downloaded), preserving the current graceful-degradation behavior rather than blanking a skill on a network error.
- A leftover staging dir from an interrupted run is cleaned by the `EXIT` trap and at start-of-run. Staging lives under `.claude/skills/` so the swap is a same-filesystem atomic rename.

Skill-level behavior (full-index re-sync, hash diff, prune of managed-but-absent skills, only touching skills in `.ic-managed.json`) is unchanged.

## Testing

Ran against the live index in a temp dir:

<details>
<summary>Results</summary>

- **Run 1 (fresh):** `26 added` — all skills + manifest downloaded.
- **Orphan test:** planted `motoko/STALE_ORPHAN.md`, tampered the stored hash to force a re-download, re-ran →
  - ✅ `motoko/SKILL.md` re-downloaded
  - ✅ `STALE_ORPHAN.md` **removed** by the atomic swap
  - ✅ no leftover `.staging-*` dirs
  - ✅ other 25 skills untouched (`25 unchanged`)
- **No-op run:** silent, no output (hash-diff still short-circuits).
- `bash -n` syntax check passes; `npm run validate` green (26 skills).

</details>

No eval file added — this is a shell-script behavior fix, not model output, so it isn't LLM-eval-able; the end-to-end test above is the verification.

### DIFF
--- a/skills/autosync-ic-skills/SKILL.md
+++ b/skills/autosync-ic-skills/SKILL.md
@@ -85,7 +85,10 @@ sync logic stays correct as it is updated upstream.
 - For each skill, compares the published `hash` against `.claude/skills/.ic-managed.json`
   (a `{ "<skill>": "<hash>" }` manifest of skills it manages) and re-downloads only the
   skills whose hash changed or are new.
-- Prunes skills it previously installed that are no longer in the index.
+- Prunes skills it previously installed that are no longer in the index, and
+  replaces a changed skill's whole directory via an atomic staging swap — so files
+  renamed or removed upstream don't linger inside a skill, and an interrupted
+  download leaves the existing copy intact.
 - Prints a one-line `added / updated / removed` summary only when something changed;
   otherwise it is silent.
 - Degrades gracefully: exits cleanly (keeping cached skills) if the network is down or
--- a/skills/autosync-ic-skills/scripts/sync-ic-skills.sh
+++ b/skills/autosync-ic-skills/scripts/sync-ic-skills.sh
@@ -17,10 +17,16 @@ MANIFEST="$DEST/.ic-managed.json"   # { "<skill>": "<hash>" } of skills this scr
 
 mkdir -p "$DEST"
 
-# --- Temp files. NEW_MANIFEST is built up as we go, then swapped in atomically. ---
+# --- Temp files. NEW_MANIFEST is built up as we go, then swapped in atomically.
+#     STAGING holds the skill dir currently being downloaded, so the trap can
+#     remove a half-written skill if the run is interrupted. ---
 TMP_INDEX="$(mktemp)"
 NEW_MANIFEST="$(mktemp)"
-trap 'rm -f "$TMP_INDEX" "$NEW_MANIFEST"' EXIT
+STAGING=""
+trap 'rm -f "$TMP_INDEX" "$NEW_MANIFEST"; [ -n "$STAGING" ] && rm -rf "$STAGING"' EXIT
+
+# Remove any staging dirs left behind by a previously interrupted run.
+rm -rf "${DEST:?}"/.staging-* 2>/dev/null || true
 
 # --- Fetch the index. On any network failure, keep cached skills and exit cleanly. ---
 if ! curl -fsSL --max-time 20 "$INDEX_URL" -o "$TMP_INDEX"; then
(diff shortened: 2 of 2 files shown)


raymondk (approved):

marc0olo:
Thanks — all three review points were valid and are addressed in 4a0610d:

1. **Path traversal via `name`** (inline) and **via `f`** (suppressed): added `is_safe_name` / `is_safe_relpath` guards that reject index-supplied skill names (no slash, no `..`) and file paths (no absolute, no `..`) before they reach any `mktemp`/`rm -rf`/`mv`/write. Applied in the prune loop, the sync loop (`name`), and the per-file download loop (`f`) — so it also hardens the pre-existing prune/download code, not just the new staging logic.
2. **rm-before-mv atomicity** (suppressed): the swap now moves the existing skill dir aside, moves the new one into place, and only then drops the old copy; if the swap fails, it restores the existing copy and keeps the old hash to retry next run. Leftover `.old-*` backups from an interrupted run are cleaned at start alongside `.staging-*`.

Tested: 12/12 guard unit cases (safe slugs/subpaths pass, `../`, `a/b`, absolute, `..` rejected); e2e run removes a planted intra-skill orphan, re-downloads the skill, leaves the other 25 untouched, no `.staging-*`/`.old-*` residue, and no-op runs stay silent. `bash -n` + `npm run validate` green.

(Realistically the index is served over HTTPS from the official site, so these are defense-in-depth against a compromised/MITM'd index — but for a script that runs `rm -rf` on remote-derived paths, worth having.)

marc0olo:
Good catch — valid, and it was a real gap in my own hardening. Fixed in 0ee8672.

The start-of-run cleanup deleted `.old-*` unconditionally, so: run interrupted mid-swap (old moved to `.old-*`, new not yet in place) → next run deletes `.old-*` before the network fetch → if that run is offline, the only remaining copy is gone. That breaks the 'existing copy stays intact' guarantee.

Now:
- Start cleanup removes only `.staging-*` (an in-progress download is always safe to discard).
- A separate **recovery** step runs *after* the index fetch (so offline runs exit first and leave `.old-*` untouched): for each `.old-<name>`, if the skill dir is **missing** it restores it (`mv` back); if the dir is **present** the backup is stale and dropped. The name derived from the backup is re-validated with `is_safe_name` before any restore.

Tested all three paths: restore (dest missing → recovered + logged), drop-stale (dest present → removed), drop-unsafe (malformed backup name → removed, not restored) — plus the earlier orphan/swap e2e still passes. `bash -n` + `npm run validate` green.

marc0olo:
Thanks — my take on the three suppressed comments:

**1. Recovery ran after the index fetch (offline gap) — VALID, fixed in 72715b9.** Good catch, and it was a suboptimal choice I made last round. Moved the `.old-*` recovery (and the guards it needs) *ahead* of the fetch, so an interrupted-mid-swap skill is now restored **even offline** (previously it stayed missing until an online run). Verified: with an unreachable index, a crash-interrupted skill is restored, the fetch still reports 'could not reach', and there's no residue.

**3. 'kept the existing copy' wording — VALID, fixed.** Reworded to 'kept any existing copy; will retry next run' since a brand-new skill has no prior copy.

**2. `rm -rf .staging-*` vs a concurrent run — VALID but I'm leaving it out of this PR, deliberately:**
- It's a **pre-existing** concurrency property, not introduced here — and this PR actually *improves* it: the old script downloaded directly into `$DEST/$name`, so two simultaneous runs interleaved writes into the *same* live skill dir (worse than racing on separate staging dirs).
- It **self-heals**: a run whose staging is deleted fails its swap, keeps/restores the existing copy, keeps the old hash, and re-syncs next session — no data loss or corruption.
- A correct fix is a single-run lock, but doing it safely needs stale-lock detection (an SIGKILL-leaked lock would otherwise deadlock a hook that runs every session), which is real complexity + new failure surface. That belongs in its own focused change, not bolted onto this orphan/swap fix.

I'd suggest a follow-up issue for 'serialize concurrent autosync runs with a stale-safe lock' rather than expanding this PR. Happy to open it if you'd like.

`bash -n` + `npm run validate` green.

marc0olo:
Re: the suppressed comment on the EXIT trap masking the exit status — this is a **false positive; no change.**

A bash EXIT trap does **not** override the script's exit status with the status of its last cleanup command. The shell exits with the status current when the trap fired, unless the trap explicitly calls `exit N` (ours doesn't). Verified on this machine:

- `set -euo pipefail` + a failing command + trap ending in a successful `rm` → exit **1** (failure preserved, not masked).
- Normal success + same trap → exit **0**.
- A trap whose last command is literally `false`, with `exit 0` → exit **0** (the trap's status is ignored).

So real failures (a `set -e` trip, a failed `mv`) already propagate non-zero, while the graceful paths (`network unreachable → exit 0`, `jq missing → exit 0`) exit 0 by design. Adding `rc=$?; …; exit "$rc"` would be unnecessary. (The masking Copilot describes is real for some other shells / non-EXIT signal traps, but not a bash EXIT trap without an explicit `exit`.)

PR #302, opened 2026-08-04
fix(canister-security): align fetchRootKey guidance with the ic_env cookie pattern

### LINKED ISSUE #266: Contradiction: fetchRootKey() banned in internet-identity, endorsed in canister-security
Two skills give opposite instructions for handling the root key in local development.

- `skills/internet-identity/SKILL.md:39` — "Never call `fetchRootKey()`" — use the `ic_env` cookie's `IC_ROOT_KEY` via the `rootKey` option to `HttpAgent.create()`, with no environment branching.
- `skills/canister-security/SKILL.md:45` — "Only call `fetchRootKey()` in local development, guarded by an environment check."

An agent reading both gets contradictory guidance for the local-dev case. The two skills should converge on one story (presumably the `ic_env` cookie / `rootKey` option pattern).

raymondk:
Closes #266.

## Problem

`canister-security` pitfall 7 told agents to call `fetchRootKey()` in local development behind an environment check, while `internet-identity` pitfall 5 bans the call outright in favor of reading `IC_ROOT_KEY` from the `ic_env` cookie and passing it via the `rootKey` option to `HttpAgent.create()`. An agent reading both skills got opposite instructions for the same local-dev scenario.

## Change

- Rewrote `canister-security` pitfall 7 to the `ic_env` cookie / `rootKey` pattern: never call `fetchRootKey()`, no environment branching, `safeGetCanisterEnv()` from `@icp-sdk/core/agent/canister-env`, with a pointer to the icp-cli skill's `references/binding-generation.md`. The MITM rationale is preserved.
- Updated the matching eval case ("Adversarial: fetchRootKey in production") whose expected behaviors encoded the old env-guard guidance.

## Eval results (changed case, with baseline)

<details>
<summary>node scripts/evaluate-skills.js canister-security --eval 6</summary>

```
━━━ Adversarial: fetchRootKey in production ━━━

  WITH skill: 5/5 passed
    ✅ Answers NO — this is not safe for production
    ✅ Explains that calling fetchRootKey() trusts whatever the replica returns, enabling a man-in-the-middle attack on mainnet
    ✅ Recommends removing the fetchRootKey() call entirely rather than wrapping it in an environment check
    ✅ Mentions passing the root key via the rootKey option to HttpAgent (from the ic_env cookie / safeGetCanisterEnv) as the correct way to support local development
    ✅ Does NOT suggest the code is acceptable as-is

  WITHOUT skill: 3/5 passed
    ✅ Answers NO — this is not safe for production
    ✅ Explains that calling fetchRootKey() trusts whatever the replica returns, enabling a man-in-the-middle attack on mainnet
    ❌ Recommends removing the fetchRootKey() call entirely rather than wrapping it in an environment check
       → The output instead recommends wrapping the call in an environment/network check, which is the opposite of the expected recommendation to remove it entirely.
    ❌ Mentions passing the root key via the rootKey option to HttpAgent (from the ic_env cookie / safeGetCanisterEnv) as the correct way to support local development
       → The output never mentions the rootKey constructor option or ic_env/safeGetCanisterEnv; it only suggests conditionally calling fetchRootKey().
    ✅ Does NOT suggest the code is acceptable as-is
```

The baseline run recommends exactly the env-guarded pattern this PR removes — a direct demonstration of the contradiction's cost.

</details>

`npm run validate` passes (warnings only, pre-existing).


raymondk:
Corrected one supporting clause: the `ic_env` cookie carries the network's root key on mainnet too (certified-assets renders `ic_root_key=<hex>` unconditionally from `ic_cdk::api::root_key()`); it equals the agent's hardcoded mainnet key, so passing it is a no-op. The previous wording claimed the cookie omits it on mainnet. The eval case checks the recommended pattern, not this mechanism detail, so results are unaffected (see #303 for the source-level evidence).

PR #303, opened 2026-08-04
fix(icp-cli): make agentOptions host conditional so builds work on custom domains

### LINKED ISSUE #267: Contradiction: HttpAgent host window.location.origin recommended by icp-cli refs, forbidden by custom-domains
- `skills/icp-cli/references/binding-generation.md:41` — presents `host: window.location.origin` as the correct agent option pattern.
- `skills/icp-cli/references/dfx-migration.md:53` — same pattern.
- `skills/custom-domains/SKILL.md:43` — "pointing `host` at it (or at `window.location.origin`) makes calls fail. ... Leave it unset, or set it explicitly to `https://icp-api.io` — never the gateway domain." (also lines 181-183)

A frontend built following the icp-cli references breaks in exactly the custom-domain scenario custom-domains warns about. The icp-cli references should at minimum qualify the pattern.

raymondk:
Closes #267.

## Problem

`skills/icp-cli/references/binding-generation.md` and `references/dfx-migration.md` presented `host: window.location.origin` as the unconditional `HttpAgent` pattern. The `custom-domains` skill (pitfall 8) correctly states that a custom domain serves only the HTTP gateway — not `/api/v2` — so a frontend built from the icp-cli pattern fails exactly when served from a custom domain. Two skills, opposite instructions.

## Change

Both references now **omit `host` entirely** and pass only `rootKey`:

```js
const agentOptions = {
  rootKey: canisterEnv?.IC_ROOT_KEY,
};
```

`@icp-sdk/core`'s default host resolution (`determineHost()`) already picks the right API endpoint everywhere:
- Known gateway origins (`ic0.app`, `icp0.io`, `localhost`, `127.0.0.1`) → page origin, so the local gateway and the Vite dev-server `/api` proxy keep working.
- Everywhere else — custom domains, `icp.net` — → `https://icp-api.io`, which is exactly what `custom-domains` prescribes.

Also corrected the `dfx-migration.md` note that scoped `IC_ROOT_KEY` to local networks: `certified-assets` renders `ic_root_key=<hex>` into the `ic_env` cookie unconditionally (sourced from `ic_cdk::api::root_key()`), so the cookie carries the network's root key on mainnet too. Passing it as `rootKey` is a no-op on mainnet (it equals the agent's hardcoded key) and replaces `fetchRootKey()` locally.

> Note: the first commit on this branch gated `host` on `IC_ROOT_KEY` presence, assuming the cookie omits the root key on mainnet. Review surfaced that the assumption was false, which would have re-broken custom domains — see the second commit and the PR comment for the source-level evidence.

Added an adversarial eval case ("agentOptions host on a custom domain") to `evaluations/icp-cli.json`.

## Eval results (new case, with baseline)

<details>
<summary>node scripts/evaluate-skills.js icp-cli --eval 28</summary>

```
━━━ Adversarial: agentOptions host on a custom domain ━━━

  WITH skill: 4/4 passed
    ✅ Omits host from agentOptions (or explicitly says to leave it unset), relying on the agent's default host resolution
    ✅ Sets rootKey from the ic_env cookie (canisterEnv?.IC_ROOT_KEY via safeGetCanisterEnv)
    ✅ Explains that the default host resolution uses the page origin on known gateway hosts (ic0.app/icp0.io/localhost) and falls back to https://icp-api.io elsewhere, so custom domains work even though they do not serve /api/v2
    ✅ Does NOT set host to window.location.origin (conditionally or unconditionally)

  WITHOUT skill: 1/4 passed
    ❌ Omits host from agentOptions (or explicitly says to leave it unset), relying on the agent's default host resolution
       → The output explicitly sets host to a hardcoded local or icp-api.io URL via a ternary rather than omitting it.
    ❌ Sets rootKey from the ic_env cookie (canisterEnv?.IC_ROOT_KEY via safeGetCanisterEnv)
       → The output uses a boolean shouldFetchRootKey flag based on env.network, not a rootKey value sourced from canisterEnv.IC_ROOT_KEY.
    ❌ Explains that the default host resolution uses the page origin on known gateway hosts (ic0.app/icp0.io/localhost) and falls back to https://icp-api.io elsewhere, so custom domains work even though they do not serve /api/v2
       → The explanation only discusses using env.network instead of window.location, with no mention of default host resolution behavior on gateway hosts or the /api/v2 fallback rationale.
    ✅ Does NOT set host to window.location.origin (conditionally or unconditionally)
```

</details>

`npm run validate` passes (warnings only, pre-existing). No existing eval case encoded the old pattern; `custom-domains` is unchanged.

### DIFF
--- a/evaluations/icp-cli.json
+++ b/evaluations/icp-cli.json
@@ -293,10 +293,10 @@
       "name": "Adversarial: agentOptions host on a custom domain",
       "prompt": "My icp-cli frontend uses @icp-sdk/bindgen bindings and safeGetCanisterEnv(). Show the agentOptions object I should pass to createActor so the same build works on the local network AND in production, including when the site is served from a custom domain. Just the agentOptions object plus a one-sentence explanation.",
       "expected_behaviors": [
+        "Omits host from agentOptions (or explicitly says to leave it unset), relying on the agent's default host resolution",
         "Sets rootKey from the ic_env cookie (canisterEnv?.IC_ROOT_KEY via safeGetCanisterEnv)",
-        "Makes host conditional: window.location.origin only when the ic_env cookie provides a root key (local network); undefined/omitted otherwise",
-        "Explains that an omitted host defaults to https://icp-api.io on mainnet, which works on custom domains where the gateway does not serve /api/v2",
-        "Does NOT set host unconditionally to window.location.origin"
+        "Explains that the default host resolution uses the page origin on known gateway hosts (ic0.app/icp0.io/localhost) and falls back to https://icp-api.io elsewhere, so custom domains work even though they do not serve /api/v2",
+        "Does NOT set host to window.location.origin (conditionally or unconditionally)"
       ]
     }
   ],
--- a/evaluations/icp-cli.json
+++ b/evaluations/icp-cli.json
@@ -288,6 +288,16 @@
         "Does NOT use a `wasm` key inside recipe.configuration — that key does not exist and fails with a cryptic strict-mode template error naming `path`",
         "Uses a version-pinned `@dfinity/prebuilt` recipe type string (e.g. @dfinity/prebuilt@v2.1.0), not the non-functional v1.0.0"
       ]
+    },
+    {
+      "name": "Adversarial: agentOptions host on a custom domain",
+      "prompt": "My icp-cli frontend uses @icp-sdk/bindgen bindings and safeGetCanisterEnv(). Show the agentOptions object I should pass to createActor so the same build works on the local network AND in production, including when the site is served from a custom domain. Just the agentOptions object plus a one-sentence explanation.",
+      "expected_behaviors": [
+        "Sets rootKey from the ic_env cookie (canisterEnv?.IC_ROOT_KEY via safeGetCanisterEnv)",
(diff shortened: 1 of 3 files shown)


raymondk:
**Revision after review:** the first commit assumed the `ic_env` cookie carries `IC_ROOT_KEY` only on local networks and used its presence to select `host: window.location.origin`. That assumption is false:

- `certified-assets` `crates/canister-core/src/asset.rs` — `render_env_cookie()` builds `payload = "ic_root_key=<hex(DER root key)>" + …` with the doc comment "The root key always comes first."
- `crates/canister-core/src/runtime.rs` — `CanisterEnv::load()` sources it from `ic_cdk::api::root_key()`, i.e. on any network, mainnet included.

So the conditional would have picked `window.location.origin` on mainnet and re-broken custom domains. The second commit drops `host` entirely; `@icp-sdk/core`'s `determineHost()` (`packages/core/src/agent/agent/http/index.ts`) resolves an omitted host to the page origin on known gateway hosts (`ic0.app`, `icp0.io`, `localhost`, `127.0.0.1`) and to `https://icp-api.io` everywhere else — including custom domains and `icp.net`. The eval case was updated to the corrected pattern and re-run (4/4 with skill, 1/4 baseline).

PR #328, opened 2026-08-05
fix(ic-dashboard): Ledger API cursor names differ; verify step uses v4

### LINKED ISSUE #294: ic-dashboard: cursor field naming and v3/v4 preference inconsistencies
1. Cursor naming: `skills/ic-dashboard/SKILL.md:45` — "Use the `next_cursor` / `previous_cursor` from v2+ responses" vs `:121` — Ledger API comment says "`next_cursor/prev_cursor`". Either the Ledger API really uses `prev_cursor` (then :45's blanket claim is wrong for that API) or one is a typo.
2. Version preference: `:26` — "Prefer v4 for canisters" yet the skill's own verification step (`:184`) uses `/api/v3/canisters?limit=1`.

raymondk:
Closes #294

Both points resolved by calling the live APIs rather than picking a side.

## 1. Cursor naming — the comment was right, the blanket rule was wrong

`:45` claimed "the `next_cursor` / `previous_cursor` from v2+ responses" as a universal rule, contradicting the Ledger API comment at `:121`. Actual response keys today:

| API | Endpoint | Keys |
|---|---|---|
| IC v4 | `/api/v4/canisters` | `data`, `next_cursor`, `previous_cursor` |
| ICRC v2 | `/api/v2/ledgers` | `data`, `next_cursor`, `previous_cursor` |
| SNS v2 | `/api/v2/snses` | `data`, `previous_cursor`, `next_cursor` |
| **Ledger v2** | `/v2/accounts` | **`accounts`**, `next_cursor`, **`prev_cursor`** |

So the Ledger API differs in **two** ways, not the one the issue reported: the back-cursor is `prev_cursor`, *and* the rows sit under `accounts` rather than `data`. Either one breaks a pagination loop written from `:45`.

`:45` and `:57` now say the names are not uniform and call out the Ledger shape; the Ledger block's own comment spells out both differences.

## 2. Version preference

`:26` says prefer v4 for canisters, but the verification step at `:184` listed them via v3. `/api/v4/canisters?limit=1` returns 200, so the verify step now uses v4 and its expected output names the real keys.

I left `:77` alone — it deliberately uses v3 for a *single* canister and says "no v4 single-canister endpoint". Confirmed that claim rather than assuming it:

```
/api/v4/canisters/ryjl3-tyaaa-aaaaa-aaaba-cai  -> 404
/api/v3/canisters/ryjl3-tyaaa-aaaaa-aaaba-cai  -> 200
```

## Evals

Seeded `evaluations/ic-dashboard.json` (it did not exist).

| Configuration | Score |
|---|---|
| With skill (this PR) | **2/2** |
| With skill (old text) | 0/2 |
| Without skill | 0/2 |

Being precise about what the 0/2 runs mean: in both, the model **declined to name the fields** and asked to fetch the endpoint rather than naming them wrongly. So this case measures that the skill supplies field names that aren't guessable, and it guards against the Ledger note being dropped again — it can only pass when the skill states the Ledger-specific names. Triggers: should-trigger 2/2, should-not-trigger 1/1.

(I decided against an eval file for this skill in #321, where the wrong prose sat beside a correct example and no case could discriminate. This one does discriminate, hence the file now.)

`npm run validate` passes: 26 skills validated, all passed.


skills/ic-dashboard/SKILL.md
```diff
@@ -42,7 +42,7 @@ Full URLs for specs and UI:
...
 2. **Choose the right API** for the data you need: IC API (canisters, subnets, NNS neurons/proposals), **Ledger API for mainnet ICP** (accounts, transactions, supply), ICRC API for **other** ICRC ledgers only (ckBTC, SNS tokens, testicp — ICRC API does not expose mainnet ICP), SNS API (SNS list, neurons, proposals), Metrics API (governance, cycles, Bitcoin, etc.).
 3. **Use the OpenAPI spec** to get exact path, query, and body schemas and response shapes; prefer the spec over hand-written docs to avoid drift.
-4. **Call over HTTPS** with `GET` (or documented method). Use the `next_cursor` / `previous_cursor` from v2+ responses to request the next or previous page.
+4. **Call over HTTPS** with `GET` (or documented method). Use the cursor fields from v2+ responses to request the next or previous page. The names are not uniform: IC, ICRC and SNS APIs return `next_cursor` / `previous_cursor`, but **Ledger API returns `next_cursor` / `prev_cursor`** and puts the rows under `accounts` rather than `data`. Read the field names from the response (or the OpenAPI spec) rather than assuming.
```
dylancm4: `/v2/accounts` returns rows under `accounts` but `/v2/transactions` returns them under `blocks`, so the key varies by endpoint rather than being a Ledger-wide accounts.

```suggestion
4. **Call over HTTPS** with `GET` (or documented method). Use the cursor fields from v2+ responses to request the next or previous page. The names are not uniform: IC, ICRC and SNS APIs return `next_cursor` / `previous_cursor`, but **Ledger API returns `next_cursor` / `prev_cursor`** and names the row array after the resource (`accounts`, `blocks`, …) rather than a uniform `data`. Read the field names from the response (or the OpenAPI spec) rather than assuming.
```

PR #335, opened 2026-08-06
fix(static-site): correct five inaccurate claims found by verifying the skill against the source

lwshang:
Every factual claim in `skills/static-site/SKILL.md` was verified against the sources it describes: [dfinity/certified-assets](https://github.com/dfinity/certified-assets) at `v0.3.3` (the version the skill pins) and [dfinity/icp-cli](https://github.com/dfinity/icp-cli) at `v1.2.0`.

The skill is accurate on nearly all of its technical detail. Five claims were wrong; four more were imprecise enough to mislead an agent. One commit per finding, each citing the files that establish the correct behavior.

## Wrong

| Commit | Was | Actually |
|---|---|---|
| `a59b12f` | a stray `.ic-assets.json5` "is uploaded as an ordinary served asset" | never uploaded — the scanner skips all dotfiles, exempting only `.well-known/` |
| `766fc7b` | `compatibility: icp-cli >= 0.3.0` | `>= 1.0.0`. The skill's own mainnet URL (`icp.net`) only became the default gateway domain in icp-cli v1.0.0; on 0.3.x it is `icp0.io`, so the documented hostname does not resolve |
| `95097cd` | canister IDs in `.icp/data/mappings/` | connected networks only; local IDs are in `.icp/cache/mappings/<env>.ids.json` |
| `7c8cefe` | legacy→static-site upgrade "panics (Cannot parse header)" | `icp deploy` aborts at the Candid compatibility check, before `install_code` — nothing installed, no state touched. Fix is `--mode reinstall`; the trap worth warning about is `--yes`, which bypasses the check and does leave a canister serving nothing |
| `0bbfc45` | "a breaking (minor/major) release reinstalls and re-uploads" | nothing reinstalls itself. The deploy upgrades in place, then the version-locked sync plugin refuses and prints the mismatch plus the command to run |

## Imprecise

- `7a5d2e3` — `ic-wasm` is a separate binary, not bundled with icp-cli, and this recipe only invokes it when `metadata:` is set (the guard sits inside `{{#if metadata}}` in the recipe template).
- `34d2fee` — the local URL's middle label is the environment name, not a literal `local`; an environment named `dev` is at `frontend.dev.localhost:8000`.
- `fcc900f` — the compressible set is wider than "text/JS/JSON/SVG/wasm" (all `text/*`, any `+json`/`+xml` suffix, `font/*` except woff/woff2), and an encoding is dropped unless it came out smaller.
- `5647dfa` — `metadata` entries take an optional `visibility`; omitting it yields a private wasm section.

## Structural

`3628f74` — pitfall 10 covered two situations that fail differently and need different commands, so it is now 10 (legacy switch) and 11 (version bump). Following pitfalls renumbered, all four body cross-references updated.

Unchanged claims were checked too, including the reserved-header list, the ~1.9 MB single-chunk limit on 4xx targets, rule precedence, the access-protection and authorize signatures, and the `v0.3.3` recipe pin.

## Notes

CI covers `skill-validator`, which could not be run locally. Its token warning (5,956 vs. a recommended 5,000) predates this branch, though these edits add ~600 tokens to it; moving the legacy/migration material into `references/` would bring it under budget.

Only `skills/static-site/SKILL.md` is touched. The two `references/` files are unreviewed — `migrating-from-asset-canister.md` is the likeliest place for the "Cannot parse header" claim corrected in `7c8cefe` to survive.


marc0olo:
Picked this up and re-verified every claim against the sources (certified-assets `v0.3.3`, icp-cli `v1.2.0`). The corrections all hold — I independently confirmed the five "Wrong" fixes plus the reserved-header list (9 total; `ETag`/`Location` are rejected on their own paths, so the list is right), the `MAX_CHUNK_SIZE = 1_900_000` limit, and the compression set.

Pushed one follow-up commit (b4a51cf) to close two spots where the *old, corrected-away* claim had survived:

**References still carried the "Cannot parse header" panic**
- `references/migrating-from-asset-canister.md` still said the in-place upgrade "panics (Cannot parse header)". Rewrote it to match the body: the plain deploy aborts at icp-cli's **pre-install Candid compatibility check** (nothing installed, running canister untouched), `--mode reinstall` skips the check because it isn't an upgrade, and `--yes` is the real trap. Backed by `deploy.rs` (check runs before `install_many`), `candid_compat.rs`, and certified-assets `docs/how-it-works.md` (state lives in `ic-stable-structures`, no deserialize-on-boot step — so there is no header to fail to parse).
- `references/legacy-asset-canister.md` quoted the same string for a *legacy-canister downgrade* (a genuinely different, serialize-on-upgrade mechanism). Reworded to the mechanism without asserting an exact error string I couldn't verify.

**Evals asserted the corrected-away behavior**
- The migration eval's first expected behavior literally required "incompatible stable-memory layouts ... panics" — it now tests the Candid-check abort.
- Added an adversarial case for the new version-bump pitfall (11).

<details>
<summary>Eval results (changed + new case, with baseline)</summary>

```
Adversarial: migrating asset-canister to static-site requires reinstall
  WITH skill:    4/4
  WITHOUT skill: 3/4   (baseline hallucinates the "Cannot parse header" panic —
                        the exact misconception this PR removes)

Adversarial: recipe version bump does not reinstall itself   (new)
  WITH skill:    4/4
  WITHOUT skill: 2/4
```
</details>

`npm run validate` passes (pre-existing warnings only). One note on the token concern in the description: the validator reports the body at **4,037 tokens**, under the 5,000 budget — so relocating the legacy/migration material isn't needed for budget reasons.

marc0olo (approved):

PR #339, opened 2026-08-07
docs(internet-identity): explain how to serve an app at more than one origin

yhabib:
# Motivation

The skill only mentioned `derivationOrigin` and `ii-alternative-origins` in Pitfall 8, to say don't use them. That is right for the `icp0.io` vs `ic0.app` case it covers, but it was the whole coverage, so an agent adding a custom domain to an II app found a prohibition and no recipe. `custom-domains` Pitfall 9 already pointed here for one. The engine console started surfacing this in dfinity/control-panel#3394 and dfinity/control-panel#3398.

# Changes

- Added a "Serving an app at more than one origin" section: pick the canister address as the derivation origin, set `derivationOrigin` on the alternative origin only, and serve `.well-known/ii-alternative-origins` from the derivation origin's canister.
- Added the `_headers` block the `@dfinity/static-site` recipe needs. The [official docs](https://docs.internetcomputer.org/building-apps/authentication/alternative-origins) say to use `.ic-assets.json5`, which is the legacy asset canister's config — static-site does not read or upload it. The file also has no extension, so it will not serve as `application/json` without a `_headers` entry, and certified-assets sets no CORS header by default.
- Scoped Pitfall 8 to the `icp0.io`/`ic0.app` case and pointed it at the new section.
- Added one eval case. Dropped a second one I drafted ("which origin gets `derivationOrigin`") because the baseline scored 3/3 on it — it tested common knowledge, not the skill.

<details>
<summary>Eval 14 — Adversarial: alternative origins on the static-site recipe (WITH 3/3 | WITHOUT 2/3)</summary>

```
━━━ Adversarial: alternative origins on the static-site recipe ━━━

  WITH skill: 3/3 passed
    ✅ Serves .well-known/ii-alternative-origins containing {"alternativeOrigins": [...]}
    ✅ Adds a _headers entry setting Content-Type: application/json AND Access-Control-Allow-Origin: * for /.well-known/ii-alternative-origins
    ✅ Does NOT create a .ic-assets.json5 file

  WITHOUT skill: 2/3 passed
    ✅ Serves .well-known/ii-alternative-origins containing {"alternativeOrigins": [...]}
    ❌ Adds a _headers entry setting Content-Type: application/json AND Access-Control-Allow-Origin: * for /.well-known/ii-alternative-origins
       → No _headers file or entry is included anywhere in the output.
    ✅ Does NOT create a .ic-assets.json5 file
```

The baseline writes the JSON file and stops, which leaves it served with the wrong content type and no CORS header.

</details>

---

Next #340, #341


marc0olo:
Pushed a small follow-up commit (8342435) folding one more fix into Pitfall 8 while we were in there.

**What:** Pitfall 8 listed only `icp0.io`/`ic0.app` as the domains II treats as equivalent, but II canonicalizes **three** official gateway domains. From `validateDerivationOrigin.ts` (`remapToLegacyDomain`):

> Compare in canonical (ic0.app) form so that the three official canister gateway domains (ic0.app, icp0.io, icp.net) are treated as one site.

and `canisterIdResolution.ts` lists `["ic0.app", "icp0.io", "icp.net", "internetcomputer.org", "localhost"]`.

**Why it belongs in this PR:** `icp.net` is now the [default domain for new frontend canisters](https://forum.dfinity.org/t/frontend-canisters-now-can-use-icp-net-instead-of-icp0-io/74901) (replacing `icp0.io`), and it is the domain this skill uses as its canonical example everywhere — including `https://<canister-id>.icp.net` as the derivation origin in the new section this PR adds. Leaving it out of Pitfall 8's "these are equivalent, don't add config" list was an internal inconsistency: the pitfall omitted the very domain the recipe recommends.

The new section itself is unaffected/correct — using `.icp.net` as the derivation origin works precisely because II canonicalizes it, and the genuine multi-origin case it covers is a custom domain, not a gateway variant.

`npm run validate` passes (warnings only). No eval change needed — this is a factual correction to an existing pitfall, and Eval 14 (added here) already covers the alternative-origins recipe.

marc0olo (approved):

PR #340, opened 2026-08-07
docs(deploy-to-cloud-engine): pin the II derivation origin at first deploy

yhabib:
# Motivation

The engine console now shows an app's canister address and names it as the origin an Internet Identity app should derive from (dfinity/control-panel#3394, dfinity/control-panel#3398). The console can only tell the user — the canisters are the user's to write, so the agent deploying them is the one that has to act. And it cannot be fixed later: a domain that has already collected sign-ins cannot be repointed at a derivation origin without orphaning every account made under it.

# Changes

- Added a subsection to Step 2 saying to pin the derivation origin on the first deploy, before the app has users.
- Pointed at the canister that **serves the frontend** as the one whose address to use, built from its canister id.
- Called out that `__META_BASE_URL` is not the answer. It is allowed to hold a custom domain, and a custom domain is the one value that must never become the derivation origin.
- Demoted `__META_MAIN_CANISTER` to the usual case rather than the rule. It is a console display flag and does not promise the canister serves HTTP.
- Noted the Caffeine exception: Caffeine owns those apps' addresses, so the address is not the app's to derive from.
- Added Pitfall 18 and one eval case. Left the mechanics in `internet-identity` and linked to it.

<details>
<summary>Eval 17 — Adversarial: II derivation origin when __META_BASE_URL is a custom domain (WITH 3/3 | WITHOUT 1/3)</summary>

```
━━━ Adversarial: II derivation origin when __META_BASE_URL is a custom domain ━━━

  WITH skill: 3/3 passed
    ✅ Gives the frontend canister address https://abcde-aaaaa-aaaaa-aaaaa-cai.icp.net as the derivation origin
    ✅ Explicitly does NOT use the custom domain https://app.example.com as the derivation origin, even though __META_BASE_URL is set to it
    ✅ Says to set it on the first deploy, before the app has users, because it cannot be applied retroactively without orphaning existing accounts

  WITHOUT skill: 1/3 passed
    ❌ Gives the frontend canister address https://abcde-aaaaa-aaaaa-aaaaa-cai.icp.net as the derivation origin
       → The output gives https://abcde-aaaaa-aaaaa-aaaaa-cai.icp0.io instead of the expected icp.net domain.
    ✅ Explicitly does NOT use the custom domain https://app.example.com as the derivation origin, even though __META_BASE_URL is set to it
    ❌ Says to set it on the first deploy, before the app has users, because it cannot be applied retroactively without orphaning existing accounts
       → The output only warns that login fails if the alternative-origins file is missing, but never states that derivationOrigin must be set before any users exist or that changing it later orphans existing accounts.
```

The case does **not** isolate the `__META_BASE_URL` correction. I ran it against the pre-fix text and that scores 3/3 too, because the rest of the section says the canister address "survives adding, changing, or dropping a custom domain", so an agent reads past the bad clause instead of following it literally. Treat this as a regression guard for the corrected guidance, not as evidence for the correction. The correction stands on the text having been self-contradictory.

I dropped a fourth behavior I had drafted ("lists the custom domain as an alternative origin") because it tests the `.well-known` mechanics this skill deliberately delegates to `internet-identity`.

</details>

---

Prev. #339

### DIFF
--- a/evaluations/deploy-to-cloud-engine.json
+++ b/evaluations/deploy-to-cloud-engine.json
@@ -1,7 +1,6 @@
 {
   "skill": "deploy-to-cloud-engine",
   "description": "Evaluation cases for the deploy-to-cloud-engine skill. Tests whether agents drive the correct cloud-engine deploy flow (link the CLI identity against the console origin, target the engine's subnet with icp deploy), use the default console origin https://opencloud.org unless the user names another, ask for the subnet id instead of guessing, set the `__META_*` environment variables so the console shows a named app with labelled canisters and an app icon (`__META_BASE_URL` + `__META_ICON_PATH` on the main canister), and avoid the documented pitfalls.",
-
   "output_evals": [
     {
       "name": "Deploy a built project to a cloud engine",
@@ -167,9 +166,18 @@
         "Sets `service:version` with an explicit version value under the recipe configuration's `metadata` list (and no git-derived entries), noting it should be bumped on each deploy",
         "Does NOT put the metadata under `settings.environment_variables` (that is for `__META_*` console variables, not wasm metadata)"
       ]
+    },
+    {
+      "name": "Adversarial: II app on an engine with a custom domain planned",
+      "prompt": "My multi-canister app is deployed to my OpenCloud cloud engine and uses Internet Identity. We will move it to a custom domain later. Which exact URL should I use as the Internet Identity derivation origin, and when do I set it? Be specific about which canister. No full deploy walkthrough.",
+      "expected_behaviors": [
+        "Names the MAIN canister (the one marked __META_MAIN_CANISTER: \"true\") as the one whose address is the derivation origin — not an arbitrary or backend canister",
+        "Gives the canister address form https://<main-canister-id>.icp.net and notes it is the same URL as __META_BASE_URL",
+        "Says to set it on the first deploy, before the app has users, because it cannot be applied retroactively without orphaning existing accounts",
+        "Does NOT recommend using the future custom domain as the derivation origin"
+      ]
     }
   ],
-
   "trigger_evals": {
     "description": "Queries to test whether the skill activates correctly. 'should_trigger' queries should cause the skill to load; 'should_not_trigger' queries should NOT activate this skill.",
     "should_trigger": [
--- a/evaluations/deploy-to-cloud-engine.json
+++ b/evaluations/deploy-to-cloud-engine.json
@@ -168,13 +168,12 @@
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


marc0olo:
Both the new subsection and Pitfall 18 call the derivation origin "the same [value/URL] as `__META_BASE_URL`". But `__META_BASE_URL` can be a custom domain (skill lines 151 and 219: "you can set it up front"), while the derivation origin must stay the **canister address** — which is exactly why the console derives it from the canister id, not the base URL:

> Unlike `defaultDomain` it never moves to a custom domain, which is what makes it the address an Internet Identity app should derive from.

So when `__META_BASE_URL` is a custom domain, "same as `__META_BASE_URL`" aims `derivationOrigin` at the custom domain — the mistake this PR is meant to prevent.

Suggest dropping the equivalence: it is the main canister's **canister address** `https://<main-canister-id>.icp.net`, which matches `__META_BASE_URL` only while that has not been pointed at a custom domain. The eval's behavior `#2` ("notes it is the same URL as `__META_BASE_URL`") carries the same conflation and would need the same softening.

yhabib:
Good catch, you're right. Fixed in 6a8f4e3.

`__META_BASE_URL` and the derivation origin agree only at first deploy, and they diverge in exactly the case the guidance is written for. Dropped the equivalence and turned it into an explicit warning, since a reader who has already pointed `__META_BASE_URL` at a domain is the one most likely to make the inference I did.

I also dropped `__META_MAIN_CANISTER` as the anchor while I was in there. It is a console display flag and does not promise the canister serves HTTP. If someone marks a backend canister, the old wording sent them to an origin the browser is never on and which cannot serve the `.well-known/ii-alternative-origins` file II fetches to validate the claim. The rule now points at the canister that serves the frontend, with the flag as the usual case and `__META_BASE_URL` called out as never the answer.

One thing worth flagging, since it cuts against the fix. I rewrote eval 17 around a setup where `__META_BASE_URL` is a custom domain, then ran it against the **old** text to confirm it catches the bug. It does not: the old text also scores 3/3. The rest of the section says the canister address "survives adding, changing, or dropping a custom domain", so an agent reads past the bad clause rather than following it literally.

So the practical blast radius was smaller than my PR description implied, and the eval is a regression guard for the corrected guidance rather than proof of this fix. The text was still self-contradictory and worth correcting on its own. Current numbers are WITH 3/3 | WITHOUT 1/3, with the delta on the `icp.net` domain and the before-you-have-users timing.

I dropped a fourth behavior I had drafted ("lists the custom domain as an alternative origin") because it tests the `.well-known` mechanics this skill deliberately delegates to `internet-identity`.

marc0olo (approved):

PR #362, opened 2026-08-25
docs(internet-identity): document the ii-app-metadata document

aterga:
## What

Documents Internet Identity's **app metadata** document (`/.well-known/ii-app-metadata`) in the `internet-identity` skill, based on [dfinity/developer-docs#347](https://github.com/dfinity/developer-docs/pull/347), the normative "App metadata" section of the [Internet Identity specification](https://docs.internetcomputer.org/references/internet-identity-spec/#app-metadata), and the implementation on `dfinity/internet-identity@main`.

Any app can now publish its own name, description, and logo for the II sign-in screens — permissionlessly, superseding the curated list II ships for a handful of known apps.

## Changes

**`skills/internet-identity/SKILL.md`**

- New section **"Showing your app's name, description, and logo on the sign-in screen"**, placed after "Serving an app at more than one origin" since the two interact:
  - the document shape, and that II reads it from the **derivation origin** (so alternative origins inherit it — no per-origin copy)
  - the field rules that decide whether the document is used at all: 40/120 code-point limits, at least one visible character, rejected control/bidi-override characters, balanced isolates, and **one bad field invalidating the whole document**
  - `logo` requirements: raster only (SVG rejected), same origin, ≤ 1 MiB, ≤ 4096 px/axis, re-encoded by II at ≤ 512 px — plus **write the URL relative**, because II may fetch the document from a sibling gateway domain of the same canister and the same-origin check runs against whichever one answered
  - the split between failures that lose the whole document and failures that cost only the logo
  - transport rules: ≤ 8 KiB, `200`, no redirects, no credentials, 10 s timeout
  - CORS headers for **both** the document and the logo, which fail differently
- Two new pitfalls in "Mistakes That Break Your Build": serving the document on the wrong origin or without CORS (13), and expecting a bad field to be dropped / confusing a rejected logo with a rejected document (14).
- Description mentions the document so queries about the sign-in screen's name and logo route here.

**Deviation from the upstream docs PR:** the docs PR configures CORS with `.ic-assets.json5`. This skill targets the `@dfinity/static-site` recipe, whose canister does not read that file (it isn't even uploaded), so the snippet uses the `_headers` file — matching the adjacent alternative-origins section and the `static-site` skill. The extension-less path also needs the bare `Content-Type:` form to get `application/json`.

**`evaluations/internet-identity.json`** — four output evals (static-site file layout, adversarial SVG + partial-validation, which origin serves the document, adversarial absolute logo URL on a gateway domain) and one trigger query.

## Evals

All four added cases run with baseline; trigger evals re-run because the description changed.

<details>
<summary>Output evals — added cases (with skill vs baseline)</summary>

```
━━━ App metadata document on the static-site recipe ━━━

  WITH skill: 5/5 passed
    ✅ Serves a JSON document at .well-known/ii-app-metadata with name, description, and/or logo fields
    ✅ Adds a _headers entry setting Access-Control-Allow-Origin: * for /.well-known/ii-app-metadata AND for the logo asset
    ✅ Sets Content-Type: application/json for /.well-known/ii-app-metadata (the file has no extension)
    ✅ Uses a raster logo (PNG/JPEG/WebP/GIF/AVIF) on the same origin as the document — not an SVG and not a cross-origin URL
    ✅ Does NOT create a .ic-assets.json5 file

  WITHOUT skill: 1/5 passed
    ❌ Serves a JSON document at .well-known/ii-app-metadata with name, description, and/or logo fields
       → The assistant produced no files at all, only a request for WebFetch/WebSearch permission.
    ❌ Adds a _headers entry setting Access-Control-Allow-Origin: * for /.well-known/ii-app-metadata AND for the logo asset
       → No _headers file or any file content was output.
    ❌ Sets Content-Type: application/json for /.well-known/ii-app-metadata (the file has no extension)
       → No headers configuration was provided since no files were output.
    ❌ Uses a raster logo (PNG/JPEG/WebP/GIF/AVIF) on the same origin as the document
       → No logo file or reference to one was provided in the output.
    ✅ Does NOT create a .ic-assets.json5 file

━━━ Adversarial: SVG logo and partial validation of ii-app-metadata ━━━
(oracle sharpened after the Copilot round)

  WITH skill: 5/5 passed
    ✅ Says no — the over-long description is a field that fails validation, and one invalid field invalidates the WHOLE document rather than being dropped on its own
    ✅ States the description limit is 120 Unicode code points, so a 200-character description fails validation
    ✅ States that image/svg+xml is not accepted — the logo must be a raster image
    ✅ Distinguishes the two failure modes: an SVG logo on its own costs only the logo, so it is the description — not the SVG — that takes the whole document down
    ✅ Does NOT claim sign-in itself breaks, and does not assert the screen necessarily shows the bare origin

  WITHOUT skill: 0/5 passed
    ❌ Says no — description failure invalidates the whole document rather than being dropped alone
       → The claim is framed throughout as unverified speculation ('likely', 'best recollection') rather than a confident, correct statement of fact.
    ❌ States the description limit is 120 Unicode code points
       → No specific numeric limit is ever given anywhere in the output.
    ❌ States that image/svg+xml is not accepted and raster formats are required
       → The rejection of SVG is presented as a speculative guess and only PNG/JPEG are mentioned, not the full accepted raster set.
    ❌ Distinguishes the two failure modes
       → The output instead claims the SVG issue 'alone being sufficient to break' all branding, conflating it with the whole-document failure.
    ❌ Does not claim sign-in breaks or that the screen necessarily shows the bare origin
       → The output asserts the bare-origin fallback without mentioning the curated-entry fallback for known apps.

━━━ Which origin serves ii-app-metadata with alternative origins ━━━

  WITH skill: 3/3 passed
    ✅ Says to publish it only on the derivation origin, because II fetches the document from the origin identities are derived for
    ✅ Explains that the alternative origins that derivation origin lists are then presented with the same name, description, and logo — no per-origin copy to keep in sync
    ✅ Does NOT tell the user to publish a copy on each alternative origin

  WITHOUT skill: 0/3 passed
    ❌ Says to publish it only on the derivation origin
       → The output instructs publishing on both origins and claims II reads from the relying party origin.
    ❌ Explains that alternative origins inherit the same metadata
       → The output instead describes two separate documents serving different purposes.
    ❌ Does NOT tell the user to publish a copy on each alternative origin
       → The output explicitly tells the user to publish the metadata file on both origins.

━━━ Adversarial: absolute logo URL on a canister gateway domain ━━━
(added for the gateway-twin gap found in review)

  WITH skill: 4/4 passed
    ✅ Says no, or flags it as a real risk — II normalizes a canister gateway origin onto ic0.app and falls back through the icp0.io / icp.net twins
    ✅ Explains the logo's same-origin check runs against whichever origin actually served the document
    ✅ States the consequence is losing the WHOLE document, not just the logo
    ✅ Recommends a relative URL such as /logo.png

  WITHOUT skill: 0/4 passed
    ❌ Says no, or flags it as a real risk
       → The output explicitly says 'Yes, that's fine' and never mentions gateway domain normalization or fallback risk.
    ❌ Explains the same-origin check runs against whichever origin served the document
       → The output only discusses CORS between logo and derivationOrigin; the mechanism is entirely absent.
    ❌ States the consequence is losing the WHOLE document
       → No mention of document-wide validation failure; implies at most a missing logo image.
    ❌ Recommends a relative URL such as /logo.png
       → The output never suggests a relative URL and instead endorses the absolute icp.net URL as fine.

━━━ Summary ━━━
    App metadata document on the static-site recipe:                  WITH 5/5 | WITHOUT 1/5
    Adversarial: SVG logo and partial validation of ii-app-metadata:  WITH 5/5 | WITHOUT 0/5
    Which origin serves ii-app-metadata with alternative origins:     WITH 3/3 | WITHOUT 0/3
    Adversarial: absolute logo URL on a canister gateway domain:      WITH 4/4 | WITHOUT 0/4
```

</details>

<details>
<summary>Trigger evals (description changed)</summary>

```
  Should trigger: 7/7 correct
    ✅ "Add login to my ICP frontend app"
    ✅ "How do I integrate Internet Identity?"
    ✅ "I need passkey authentication for my dapp"
    ✅ "Set up sign-in with Internet Identity"
    ✅ "How does delegation work with II?"
    ✅ "Add auth to my canister frontend"
    ✅ "How do I get my app's name and logo to show on the Internet Identity sign-in screen?"

  Should NOT trigger: 6/6 correct
    ✅ "Connect a wallet to my dapp"
    ✅ "How do I deploy my canister?"
    ✅ "Implement ICRC-1 token transfers"
    ✅ "Set up inter-canister calls"
    ✅ "How does stable memory work?"
    ✅ "Add a frontend to my canister"

  Trigger evals: should-trigger 7/7 | should-not-trigger 6/6
```

</details>

Evals 15 and 16 were re-checked after the logo-bullet edit in 05fbf01 (both still 5/5). The other pre-existing cases were not re-run — none of them cover content this PR touched.

`npm run validate` passes — 29 skills, no errors. The skill trips the validator's "body > 500 lines" advisory (547 lines) alongside the pre-existing token advisory; per `.claude/CLAUDE.md` warnings are acceptable, and the next addition to this skill should go in `references/`.

## Review rounds

**Copilot (fixed in 1bc611e)** — three findings, all valid:

- Pitfall 13 conflated the two CORS failure modes. An unreadable **document** means none of your metadata is used; an unreadable **logo** costs only the logo.
- "The app is shown with no metadata" skipped the curated fallback. Now: none of *your document* is applied, then the fallback order.
- The adversarial eval's oracle did not test the distinction the section introduces. Sharpened: 4/4 vs 2/4 → **5/5 vs 0/5**.

Pitfall 14 also now splits `logo` across both modes: a cross-origin URL fails *document* validation and takes the whole document down, while an SVG, an oversized image, or one that cannot be fetched costs only the logo.

**@marc0olo (fixed in 05fbf01)** — two must-fixes, both verified against `dfinity/internet-identity@main` before applying:

- **The spec link 301'd to the docs homepage**, and the validator's link check follows redirects and sees `200`, so CI could not catch it. Switched to the `docs.internetcomputer.org/<path>` form used by every other IC-docs link under `skills/`, with the now-published `#app-metadata` anchor.
- **The gateway-twin fallback was uncovered**, which made an absolute `logo` URL a whole-document failure under this skill's own recommended derivation origin: `appMetadataOrigins` tries the `ic0.app` / `icp0.io` / `icp.net` twins in turn, and `validateLogoUrl` checks same-origin against whichever answered, returning the document-level `INVALID`. The logo bullet now says to write the URL relative and why; pitfall 14 no longer reads as being only about third-party hosts. New adversarial eval: **4/4 with the skill, 0/4 baseline** — the baseline opens with "Yes, that's fine".

Non-blocking nits from that review are answered in the thread: `Content-Type: application/json` on the document is hygiene rather than an II requirement (`fetchCapped` sends it as `Accept` and never inspects the response), left as-is since the text does not mis-state it; `credentials: omit` and the 10 s timeout are implementation-only but load-bearing when debugging a fetch.


skills/internet-identity/SKILL.md
```diff
@@ -57,6 +57,10 @@ Internet Identity (II) is the Internet Computer's native authentication system.
     - `verified_email` is the same email as `email`, but only present when the source OpenID provider (e.g., Google) marked it as verified and II surfaced that signal through.
     Use `verified_email` for any access gating (admin allowlists, capability checks). Use `email` only for soft uses like contact info or mailing lists. Request both for fallback behaviour: both are returned with the same value when the source provider marked the email as verified, only `email` when it didn't.
 
+13. **Serving `/.well-known/ii-app-metadata` on the wrong origin, or without CORS.** II reads app metadata from the origin identities are derived for — your validated `derivationOrigin` when the request sets one, the request's own origin otherwise. A document published only on the alternative origin the user visits is never fetched. Both the document *and* the logo it points at are read cross-origin, so both need `Access-Control-Allow-Origin`; without it the sign-in screen silently falls back to showing the bare origin. See "Showing your app's name, description, and logo on the sign-in screen".
```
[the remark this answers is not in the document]
aterga: Correct — fixed in 1bc611e. Pitfall 13 now states the two CORS failures separately: an unreadable document means none of your metadata is used (curated entry if II ships one for your app, origin alone otherwise), while an unreadable logo costs only the logo.

---

skills/internet-identity/SKILL.md
```diff
@@ -183,6 +187,47 @@ Do **not** reach for `.ic-assets.json5` — that is the legacy asset canister's
...
+- Each field must contain at least one visible character. A field holding only whitespace or invisible characters is rejected, not treated as absent.
+- Control characters (other than the ASCII whitespace `\t`, `\n`, `\v`, `\f`, `\r`), `U+FEFF`, and the bidirectional embeddings and overrides `U+202A`–`U+202E` are rejected: they can make rendered text read differently from what it contains. The characters mixed-direction and non-Latin names legitimately need are accepted — the marks `U+200E`, `U+200F`, `U+061C`, the isolates `U+2066`–`U+2069`, and the zero-width characters `U+200B`–`U+200D` — but isolates must be **balanced**: a field must close every isolate it opens and close none it did not open.
+- **One bad field invalidates the whole document**, which is then ignored; the offending field is not dropped on its own. The app is shown with no metadata rather than half of it. II logs which field is at fault to the browser console — check the console on the sign-in screen when metadata does not appear. A document carrying no field II recognises is likewise ignored; a valid one replaces the curated fallback entry wholesale.
```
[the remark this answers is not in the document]
aterga: Fixed in 1bc611e. The bullet now says none of *your document* is applied, then names the fallback order (curated entry for a known app, origin alone otherwise). Pitfall 14 was reworded the same way, and it now splits `logo` across the two modes explicitly: a cross-origin URL fails document validation and takes the whole document down, while an SVG, an oversized image, or one that cannot be fetched costs only the logo.

---

evaluations/internet-identity.json
```diff
@@ -137,6 +137,36 @@
...
+      ]
+    },
+    {
+      "name": "Adversarial: SVG logo and partial validation of ii-app-metadata",
+      "prompt": "My /.well-known/ii-app-metadata has \"logo\": \"/logo.svg\" and a 200-character description. II will just skip the description that's too long and still show my SVG logo and name, right? Answer briefly.",
+      "expected_behaviors": [
+        "Says no — one invalid field invalidates the WHOLE document, so II ignores all of the metadata rather than dropping only the bad field",
+        "States the description limit is 120 Unicode code points, so a 200-character description fails validation",
+        "States that image/svg+xml is not accepted — the logo must be a raster image (image/png, image/jpeg, image/webp, image/gif, or image/avif)",
+        "Does NOT claim sign-in itself breaks — invalid metadata only falls back to showing the origin"
```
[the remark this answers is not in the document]
aterga: Good catch — both distinctions are now in the oracle (1bc611e). The prompt asks what each of the two problems costs, and the expectations require the model to attribute the whole-document loss to the over-long description, note that an SVG logo alone would cost only the logo, and not assert the bare-origin fallback unconditionally.

Re-ran with baseline: **5/5 with the skill, 0/5 without** (previously 4/4 vs 2/4), so the sharpened oracle also made the case a much stronger regression test. Updated results are in the PR description.

---

skills/internet-identity/SKILL.md
```diff
@@ -183,6 +187,47 @@ Do **not** reach for `.ic-assets.json5` — that is the legacy asset canister's
...
+
+/logo.png
+  Access-Control-Allow-Origin: *
+```
+
+`.well-known/` is uploaded automatically, but this file has no extension either, so set its media type with the bare `Content-Type:` form. As above, `.ic-assets.json5` is the legacy asset canister's config and is not read by this recipe.
+
+Missing, unreachable, or invalid metadata never blocks sign-in: the screen falls back to the curated entry, and to your origin alone otherwise. And because the file is under the sole control of the origin serving it, publishing it verifies nothing about your app — II keeps displaying the origin alongside whatever you provide, since the origin is the part users can actually check.
+
+The normative rules, including a JSON Schema you can validate the document against in CI, are in the **App metadata** section of the [Internet Identity specification](https://internetcomputer.org/docs/references/internet-identity-spec). The schema expresses every rule above except isolate balancing.
```
marc0olo: **This link resolves to the docs homepage, not the spec — and CI can't catch it.**

```
$ curl -sIL -o /dev/null -w '%{url_effective} %{http_code}\n' \
    https://internetcomputer.org/docs/references/internet-identity-spec
https://docs.internetcomputer.org/ 200
```

It 301s to the site root. The validator's link check follows the redirect, sees `200`, and passes — so this is invisible to the pipeline. It's also the only `internetcomputer.org/docs/...` link in the repo; every other IC-docs link under `skills/` uses the `docs.internetcomputer.org/<path>` form (zero occurrences of the other shape on `main`).

**And the anchor note in the PR description is stale — the section is published.** Both `id="app-metadata"` and `id="app-metadata-schema"` are live on the page now, so the fragment isn't dead:

```
$ curl -sL https://docs.internetcomputer.org/references/internet-identity-spec/ \
    | grep -o 'id="app-metadata[a-z-]*"' | sort -u
id="app-metadata"
id="app-metadata-schema"
```

Suggest linking straight to the section (or `#app-metadata-schema`, which is arguably the better target given the sentence is about validating in CI) and dropping the Notes paragraph from the description:

```suggestion
The normative rules, including a JSON Schema you can validate the document against in CI, are in the **App metadata** section of the [Internet Identity specification](https://docs.internetcomputer.org/references/internet-identity-spec/#app-metadata). The schema expresses every rule above except isolate balancing.
```

skills/internet-identity/SKILL.md
```diff
@@ -183,6 +187,47 @@ Do **not** reach for `.ic-assets.json5` — that is the legacy asset canister's
...
+- **One bad field invalidates the whole document**, which is then ignored; the offending field is not dropped on its own. None of your document is applied rather than half of it; II then falls back to its curated entry if it ships one for your app, and to your origin alone otherwise. II logs which field is at fault to the browser console — check the console on the sign-in screen when metadata does not appear. A document carrying no field II recognises is likewise ignored; a valid one replaces the curated fallback entry wholesale.
+- `logo` must be a **raster** image URL on the **same origin** as the document (relative URLs resolve against it), served as `image/png`, `image/jpeg`, `image/webp`, `image/gif`, or `image/avif`, at most 1 MiB, and at most 4096 pixels per axis. `image/svg+xml` is rejected — serve a rasterised copy of a vector logo. II downloads the image (it is never hotlinked), redraws it at up to 512 pixels on its longest side (an animated image is flattened to its first frame), and renders that copy, so a roughly square PNG or WebP of about 512 pixels is the right thing to ship.
```
marc0olo: **The gateway-twin fallback isn't covered here, and it turns an absolute `logo` URL into a whole-document failure — under this skill's own recommended setup.**

The section says II fetches from the derivation origin, full stop. The implementation does more: a canister gateway origin is first remapped onto `ic0.app` (`remapToLegacyDomain` — principals must not depend on which gateway the user arrived through), and `appMetadataOrigins` then falls back through the `icp0.io` / `icp.net` twins, stopping at the first origin that answers for the app. `ii-spec.mdx` covers this in the paragraph beginning *"That origin is normalized onto `ic0.app` …"*.

The catch is that `validateLogoUrl` runs its same-origin check against **whichever gateway origin actually answered**:

```ts
if (url.origin !== new URL(origin).origin) {
  return reject("logo", "must be on the same origin as the document");
}
```

and `reject` returns the document-level `INVALID`. So with the derivation origin recommended in the alternative-origins section directly above (`https://<canister-id>.icp.net`), a developer who writes

```json
{ "logo": "https://<canister-id>.icp.net/logo.png" }
```

loses the **entire document** whenever II fetched it from `<canister-id>.ic0.app` — same canister, different origin. The text does say "relative URLs resolve against it", but never that relative is what you should *use*, and pitfall 14's cross-origin warning reads as being about third-party CDNs rather than your own canister on a sibling gateway domain.

One clause on this bullet would close it, e.g.:

> Use a **relative** URL: II may fetch the document from a sibling gateway domain of the same canister (`ic0.app` / `icp0.io` / `icp.net`), and an absolute URL pinned to one of them would then be cross-origin and invalidate the whole document.

This is the only substantive content gap I found.

marc0olo (commented):
Reviewed this against two sources rather than the one in the description: the normative `docs/ii-spec.mdx` §App metadata, **and** the implementation in `src/frontend/src/lib/utils/appMetadata.ts` / `lib/stores/app-metadata.store.ts` on `dfinity/internet-identity@main`.

**Verdict: sound, and accurate to an unusual degree.** Every numeric and behavioural claim checks out against the code — the 40/120 code-point limits counted pre-normalization (`validateTextField` measures `[...value].length` *before* the `\s+` collapse), the exact forbidden-character class, `hasBalancedIsolates`, the all-or-nothing `INVALID` short-circuit, `APP_LOGO_CONTENT_TYPES`, `MAX_APP_LOGO_SIZE`/`MAX_APP_LOGO_DIMENSION`/`APP_LOGO_RENDER_SIZE`, `MAX_APP_METADATA_SIZE`, `redirect: "error"`, `credentials: "omit"`, and `APP_METADATA_FETCH_TIMEOUT_MILLIS`. The `_headers` deviation from the docs PR is the right call for this repo — `static-site` pitfalls 1, 6, and 12 all back it, and `Access-Control-Allow-Origin: *` is valid here precisely *because* II omits credentials. The three Copilot findings were all real and correctly fixed in 1bc611e.

### Worth calling out: the code check mattered

The document-vs-logo split this PR asserts is **not** settled by the spec text. `ii-spec.mdx` puts the same-origin requirement in the *same* bullet as the content-type/size/dimension requirements, then says a logo that "does not meet the asset requirements above costs the application only its logo" — which reads the other way. It also claims the published JSON Schema expresses every rule except isolate balancing, while the schema's `logo` is just `{type: string, minLength: 1}`, implying same-origin *isn't* document validation.

The implementation settles it in this PR's favour: `validateLogoUrl` returns the document-level `INVALID` sentinel for a cross-origin URL, and `INVALID` short-circuits the whole document in `fetchAppMetadataFrom`, whereas `fetchLogoObjectUrl` returns `undefined` and only drops `metadata.logo`. So the skill is right and the spec prose is the weaker of the two — probably worth a separate nit upstream on `dfinity/internet-identity`.

### Two must-fixes and one content gap

See the inline comments on lines 229 and 210. Summary:

1. The spec link 301s to the docs **homepage**, not the spec — and the validator's link check follows the redirect, sees 200, and passes, so CI can't catch it.
2. The "no anchor yet" note in the description is stale; `#app-metadata` is live.
3. The gateway-twin fallback isn't covered, which makes an absolute `logo` URL a trap under this skill's *own* recommended derivation-origin pattern.

### Non-blocking nits

- **`Content-Type: application/json` on the document is hygiene, not a requirement.** `fetchCapped` sends `application/json` as an `Accept` header only and never inspects the response's content-type; the body is decoded and `JSON.parse`d regardless. Only the *logo*'s content-type is enforced. Setting it is still right and matches the block above — but that block says "II needs both", and a reader may carry the "needs" over. Only worth touching if you're editing that paragraph anyway.
- **`credentials: omit` and the 10 s timeout live only in the implementation**, not in the normative §App metadata. Fine to document; just the likeliest claims to drift.
- **Size advisory.** The skill goes 510 → 555 lines and now trips the >500-line warning alongside the pre-existing token warning. Warnings are acceptable per `.claude/CLAUDE.md` and the check is green — flagging only as a signal that the *next* addition to this skill probably belongs in `references/`.

### Evals

Structurally valid, 14 → 17 output evals, `should_trigger` 6 → 7. The adversarial case is a genuinely strong regression test after the Copilot round, and its oracle is correct against the implementation: an SVG logo passes `validateLogoUrl` (it's same-origin) and fails only at the content-type check in `fetchLogoObjectUrl`, so it really does cost only the logo — exactly what the oracle demands. Skipping a re-run of the untouched cases is correct per `.claude/CLAUDE.md`.

Everything other than the three items above is approve-worthy as written.

skills/internet-identity/SKILL.md
aterga: Confirmed and applied in 05fbf01 — took the suggestion verbatim (`#app-metadata`, matching the "App metadata section" the sentence names). Reproduced the 301 to the docs homepage, and both anchors are live, so the stale Notes paragraph is dropped from the description too.

Worth recording why CI missed it: the validator's link check follows redirects and only asserts the final status, so a 301 to a valid page passes. `--skip links` in `npm run validate` is beside the point here — even with link checks enabled this link is green. The repo-convention half of your point is the durable fix: `docs.internetcomputer.org/<path>` was the shape used by every other IC-docs link under `skills/`, and this was the one outlier.

---

skills/internet-identity/SKILL.md
aterga: Verified against `main` and fixed in 05fbf01. `fetchAppMetadataFrom(candidate)` passes its own `candidate` origin into `validateLogoUrl(logo, origin)`, so the same-origin check really does run against whichever twin answered, and `reject` → `INVALID` short-circuits alongside the name/description checks. Confirmed the trap is reachable exactly under the derivation origin this skill recommends.

The logo bullet now says to write the URL relative and gives the mechanism; pitfall 14 no longer reads as being only about third-party hosts. Added an adversarial eval for it — **4/4 with the skill, 0/4 baseline**, and the baseline answer opens with "Yes, that's fine", which is the failure mode you predicted. Evals 15 and 16 re-checked after the edit, both still pass.

On your nits: agreed that `Content-Type: application/json` on the document is hygiene rather than an II requirement — `fetchCapped` sends it as `Accept` and never inspects the response — but I left that paragraph alone rather than spending lines on a clarification the text doesn't actually mis-state, given the size signal you raised. Same reasoning for `credentials: omit` and the 10 s timeout: documented, drift-prone, but load-bearing for someone debugging a metadata fetch. And noted on `references/` for the next addition to this skill.

---

skills/internet-identity/SKILL.md
```diff
@@ -57,6 +57,10 @@ Internet Identity (II) is the Internet Computer's native authentication system.
...
+
+14. **Assuming a bad field in `ii-app-metadata` is just dropped, or confusing a rejected logo with a rejected document.** One field that fails validation invalidates the **whole document**: none of your metadata is applied, not just the offending field (II then falls back to its curated entry if it ships one for your app, and to your origin alone otherwise). `name` is capped at 40 Unicode code points and `description` at 120, counted on the value as served. `logo` straddles the two failure modes — a URL that is not on the **same origin** as the document fails document validation and takes the whole document down with it — and that includes your own canister on a sibling gateway domain, since II may fetch the document from any of `ic0.app`, `icp0.io`, or `icp.net` (write the URL relative), while an SVG (`image/svg+xml` is not accepted; serve a raster copy), an oversized image, or one that cannot be fetched or decoded costs you the logo alone.
```
marc0olo: **Punctuation nit — the em-dash pair now closes one clause too early.**

The gateway sentence landed *after* the closing dash, so the dashes no longer bracket failure-mode A, and `while` — which contrasts mode B against mode A — ends up separated from its counterpart by the entire gateway aside:

> `logo` straddles the two failure modes — [**mode A**] — and that includes … gateway domain …, while [**mode B**] costs you the logo alone.

Moving the closing dash to after `(write the URL relative)` keeps all of mode A inside the dashes and puts `while [mode B]` directly against it:

```suggestion
14. **Assuming a bad field in `ii-app-metadata` is just dropped, or confusing a rejected logo with a rejected document.** One field that fails validation invalidates the **whole document**: none of your metadata is applied, not just the offending field (II then falls back to its curated entry if it ships one for your app, and to your origin alone otherwise). `name` is capped at 40 Unicode code points and `description` at 120, counted on the value as served. `logo` straddles the two failure modes — a URL that is not on the **same origin** as the document fails document validation and takes the whole document down with it, and that includes your own canister on a sibling gateway domain, since II may fetch the document from any of `ic0.app`, `icp0.io`, or `icp.net` (write the URL relative) — while an SVG (`image/svg+xml` is not accepted; serve a raster copy), an oversized image, or one that cannot be fetched or decoded costs you the logo alone.
```

Purely readability, and only worth doing because this pitfall is the densest sentence in the skill and it is agent-facing. No objection to merging as-is if you'd rather not respin.

marc0olo (commented):
Re-reviewed `05fbf01`. **All three items addressed, and the mechanism is stated accurately.** I re-checked the new prose against `dfinity/internet-identity@main` rather than the description:

- **Spec link** → `docs.internetcomputer.org/references/internet-identity-spec/#app-metadata`, verified 200 with the anchor live. Stale Notes paragraph gone.
- **Gateway twins** — "normalizes onto `ic0.app` and tries its `icp0.io` and `icp.net` twins **in turn**" matches `GATEWAY_DOMAINS = ["ic0.app", "icp0.io", "icp.net"]` exactly, order included; scoping it to a *canister gateway* origin is right, since `gatewayOriginTwins` returns `[]` for custom domains and localhost; and "the same-origin check runs against whichever one answered" is right because the loop hands each `candidate` to `fetchAppMetadataFrom`, which passes that same origin into `validateLogoUrl`. Fixed in both places it needed to be — the `logo` bullet and pitfall 14.
- **New adversarial eval** — 18 output evals, JSON valid, and oracle 1 hedging with "says no, **or flags it as a real risk**" is the right calibration, since whether it actually breaks depends on which gateway answers first. A baseline that opens with "Yes, that's fine" is exactly the failure the case exists to catch.

`npm run validate` on the branch: 29 skills, all passed, warnings only. Re-checking evals 15 and 16 was the right call, and I did not re-run them — the logo-bullet edit is purely additive and can only reinforce both oracles, so I'd have skipped it too.

One punctuation nit inline, non-blocking. **Otherwise this is ready to merge.**

skills/internet-identity/SKILL.md
aterga: Taken in a644abd — you're right that the pair closed early and stranded `while` from the mode it contrasts against. Applied as a pure punctuation move (dash after "with it" → comma; comma after "(write the URL relative)" → dash), so the diff is one line and the wording is unchanged.

No eval re-run for this one: nothing semantic changed, and the only case covering pitfall 14 keys on the whole-document-vs-logo distinction, which the dashes don't carry. `npm run validate` still passes.

---

marc0olo (approved):

Issue #367, opened 2026-08-26
icp-cli: agents use 'candid:' instead of 'didFile:' in the bindgen Vite plugin config


marc0olo:
Surfaced while running evals for the #364 PR. **Pre-existing — not introduced by that PR** (verified below), so filing separately rather than widening its scope.

## Symptom

`evaluations/icp-cli.json` case 15 ("Full-stack Motoko config artifacts") fails one behavior with the skill loaded:

```
❌ Uses @icp-sdk/bindgen (>= 0.4.0) Vite plugin with a didFile path pointing to the committed .did file
   → The bindgen block uses a 'candid' key instead of the required 'didFile' key,
     and no version is specified for @icp-sdk/bindgen.
```

The model writes the Vite plugin as `icpBindgen({ candid: "...", outDir: "..." })` instead of `didFile:`. That config is silently wrong — bindgen ignores the unknown key and has no `.did` path.

## Confirmed pre-existing

Case 15 was re-run against the unmodified skill content (whole change set stashed):

```
BEFORE:  ❌ ... configures the path via a 'candid' key instead of the expected 'didFile' field
AFTER:   ❌ ... uses a 'candid' key instead of the required 'didFile' key
```

Identical failure, identical reason, 5/6 both ways. The only line the #364 PR touches on that case is the version number inside the expectation string.

## Root cause: two different keys for the same concept

The skill uses **two** key names for "path to the `.did` file", and the wrong one dominates by weight:

| Key | Where it belongs | Occurrences in `skills/icp-cli/` |
|-----|------------------|-------------------------------|
| `candid:` | `icp.yaml`, under `recipe.configuration` | `SKILL.md:124, 134, 149, 369`, `dfx-migration.md:102` |
| `didFile:` | `vite.config.js`, inside `icpBindgen()` | `binding-generation.md:16, 20`, `dev-server.md:62` — and **not once in `SKILL.md`** except one prose aside at `:163` |

An agent reading `SKILL.md` — the entry point — sees `candid:` presented four times as *the* key for the `.did` path and `didFile` essentially never. `binding-generation.md:91` then compounds it by discussing `candid:` in the same paragraph as bindgen:

> …commit it to the repo, and specify `candid:` in the recipe config. If `candid` is omitted, the recipe auto-generates the `.did`…

Nothing anywhere states that these are two distinct keys in two distinct files.

## Suggested fix

[...]

Issue #366, opened 2026-08-26
Add a version-coherence guard so skills cannot contradict each other on dependency versions


marc0olo:
Version drift between skills is currently found only by manual audit. #297, #298 and #301 are all instances of the same missing check, and #364 is a case where a skill's install command contradicted its own stated prerequisites for months without anything noticing.

## Evidence: a ~40-line extractor finds this today

Regexing version constraints out of `skills/**/*.md` and grouping by value:

```
CONFLICT  ic-cdk (cargo)
    0.19   <- canister-security, certified-variables, ckbtc, https-outcalls,
              icrc-ledger, multi-canister, sns
    0.20   <- evm-rpc, vetkeys

CONFLICT  mo:core
    2.1.0  <- multi-canister
    2.5.0  <- migrating-motoko-actors, troubleshooting-motoko-migrations, writing-motoko

CONFLICT  moc (toolchain)
    0.14.2 <- cloud-engine-canisters
    1.6.0  <- internet-identity
    1.7.0  <- mops-cli, stable-memory, writing-motoko
    1.11.2 <- migrating-motoko-actors, troubleshooting-motoko-migrations, writing-motoko
    1.12.0 <- mops-cli

CONFLICT  @icp-sdk/core
    ^5.0.0 <- icp-cli
    ^5.4   <- vetkeys
```

That reproduces #301, #297 and #298 automatically.

## Design note: uniformity is the wrong rule

#333 ("state each skill's real `mo:core` minimum instead of one repo-wide pin") exists precisely because these numbers *should* differ in some places. A blunt "all versions must match" check would fight that PR. Three distinct kinds need three rules:

| Kind | Example | Rule |
|---|---|---|
| **Floor** | `moc >= 1.6.0` in `compatibility:` | May legitimately differ per skill. Flag only when below a known-bad version. |
| **Pin** | `@icp-sdk/core@^5.0.0` in an install command | Must agree repo-wide. |
| **Unpinned** | `npm i @icp-sdk/core` | Always a defect in an install command. |

The third rule alone would have caught #364 at authoring time.

### Tier 1 — per-PR, offline

Extend `scripts/check-project.js` (already wired into `npm run validate` and CI, currently 76 lines doing metadata/eval checks) or add a sibling `scripts/check-versions.js` called from the same entry point.

- Flag any `npm install` / `npm i` / `mops add` / cargo line naming a shared dependency with no version constraint.
- Flag pin disagreement against a declared manifest.
[...]

Issue #365, opened 2026-08-26
Follow-up: move skills to @icp-sdk/core v6 once the dependency chain supports it


marc0olo:
Follow-up to #364, which pins every skill to `@icp-sdk/core@^5`. This issue tracks flipping them to `^6` — **do not action until the chain below is green.**

## Why we are on `^5`

`@icp-sdk/core` **6.1.0** is published (6.0.0 on 2026-07-06, 6.1.0 on 2026-08-03), but nothing downstream accepts it:

| Package | Latest | Requires core |
|---|---|---|
| `@icp-sdk/signer` | 5.6.3 | `^5` (peer) |
| `@icp-sdk/auth` | 8.0.3 | `^5` (peer) |
| `@icp-sdk/vetkeys` | 0.5.0 | `^5.4.0` (dep) |
| `@dfinity/oisy-wallet-signer` | 6.0.0 | `^4` (peer — see #364) |
| `@icp-sdk/bindgen` | 0.4.0 | no core dependency |

Installing core 6 alongside `@icp-sdk/auth` is a hard failure:

```
npm error code ERESOLVE
npm error Found: @icp-sdk/core@6.1.0
npm error Could not resolve dependency:
npm error peer @icp-sdk/core@"^5" from @icp-sdk/auth@8.0.3
```

Forcing it with `--legacy-peer-deps` is worse — it installs **two copies** of core (5.4.0 nested for vetKeys, 6.1.0 at root), so `Principal`/`HttpAgent` class identities no longer match across the boundary vetKeys passes them over.

## What core 6 actually broke

Not documented in the changelog (it lists only feats). From [icp-js-signer#39](https://github.com/dfinity/icp-js-signer/pull/39), which fixed them:

- **`Agent.readState`** — first parameter is now `InputTargetPrincipal` (`{ canisterId } | { subnetId }`), was `Principal | string`
- **`DerEncodedPublicKey`** narrowed to `Uint8Array<ArrayBuffer>`
- **`Delegation`** gains an optional 4th constructor arg (`permissions`) — additive, not breaking

The documented API surface our skills use is **unaffected**: core 6.1.0 still exports `./agent/canister-env`, `./principal`, `safeGetCanisterEnv`, `HttpAgent.create`/`createSync`. This is purely a dependency-resolution blocker.

## Upstream chain (in order)

[...]

PR #368, opened 2026-08-26
fix: align all skills on @icp-sdk/core ^5 and pin the dependencies that drift off it

### LINKED ISSUE #364: Cross-skill version coherence: unpinned @icp-sdk/core installs resolve to incompatible majors
## Problem

`skills/wallet-integration/SKILL.md:60` installs nothing pinned:

```bash
npm i @dfinity/oisy-wallet-signer @dfinity/utils @dfinity/zod-schemas @icp-sdk/canisters @icp-sdk/core zod
```

Run today, that resolves to:

```
add @icp-sdk/core 4.2.3        <- two majors below every other skill
add @icp-sdk/canisters 3.1.0   <- downgraded from 3.6.0
add @dfinity/utils 4.0.3       <- downgraded from 4.2.1
```

The same skill's **Prerequisites** section (`:55-57`) states `@icp-sdk/core (>= 5.0.0)`, `@dfinity/utils (>= 4.2.0)`, `@icp-sdk/canisters (>= 3.5.0)`. The install command violates all three of its own stated constraints.

## Cause

`@dfinity/oisy-wallet-signer` **6.0.0 regressed its peer dependencies backwards**:

| oisy | peer `@icp-sdk/core` | peer `@dfinity/utils` | peer `@icp-sdk/canisters` |
|------|------|------|------|
| 4.1.0 | `^5` | `^4.1` | `^3.2` |
| 4.1.3 | `^5` | `^4.2.1` | `^3.2` |
| **6.0.0** (latest) | **`^4`** | **`~4.0.3`** | **`~3.1.0`** |

The skill was written correctly against oisy 4.1.x. Because the install line is unpinned it now takes 6.0.0, which drags the whole tree back to core 4.

## Impact: two skills that cannot be used in the same project

- `internet-identity` → `@icp-sdk/auth` 8.0.3 → peer `@icp-sdk/core@^5`
- `wallet-integration` → oisy 6.0.0 → peer `@icp-sdk/core@^4`

Any app doing II login **and** wallet integration — a very common combination — cannot produce a valid install tree. Forcing core `^5` with the current line fails:

```
npm error code ERESOLVE
npm error Found: @icp-sdk/core@5.4.0
npm error Could not resolve dependency:
npm error peer @icp-sdk/core@"^4" from @dfinity/utils@4.0.3
```

## The whole surface *can* be coherent at core 5

Pinning oisy below its regressed major resolves everything together:

```bash
npm i '@dfinity/oisy-wallet-signer@^4.1.3' @dfinity/utils @dfinity/zod-schemas \
      @icp-sdk/canisters '@icp-sdk/core@^5' @icp-sdk/auth zod
```
```
add @icp-sdk/core 5.4.0       add @icp-sdk/auth 8.0.3      add @icp-sdk/signer 5.6.3
add @icp-sdk/canisters 3.6.0  add @dfinity/utils 4.2.1     add @dfinity/oisy-wallet-signer 4.1.3
```

vetKeys and bindgen sit on the same tree: `@icp-sdk/vetkeys` 0.5.0 depends on core `^5.4.0`, and bindgen's generated output type-checks clean (strict, `skipLibCheck: false`) against both core 5.4.0 and 6.1.0.

## Also wrong, same class

`skills/icp-cli/references/binding-generation.md:89`:

> `@icp-sdk/core` starts at version 5.x — there is no 0.x or 1.x release. Do not guess a lower version.

The published line is `1.0.0-beta.0…5` → `4.0.0…4.2.3` (stable) → `5.x` → `6.x`. So it does not start at 5, 1.x prereleases exist, and — the part that matters — **6.x is published but nothing else supports it yet**, so the sentence invites someone to take `latest` and hit `ERESOLVE` against `@icp-sdk/auth`. (The move to `^6` is tracked separately in #365 — do not action it here.)

## Rule to adopt

1. **Pin every `@icp-sdk/core` mention explicitly to `^5`.** Never leave it unpinned in an install command.
2. **Pin `@dfinity/oisy-wallet-signer` to `^4.1.3`** (below the regressed 6.0.0) until it supports core `^5` again.
3. Say *why* `^5` and not `latest`, so the next person doesn't "helpfully" bump it.

## Scope

- `skills/wallet-integration/SKILL.md` — pin the install line; reconcile with Prerequisites
- `skills/icp-cli/references/binding-generation.md` — pin, and fix the `:89` sentence
- `skills/icp-cli/references/dfx-migration.md` — already pinned `^5.0.0`; bump bindgen floor to `^0.4.0`
- `skills/icp-cli/SKILL.md:115` — same core wording
- `skills/vetkeys/SKILL.md:25` — already `^5.4`; confirm and keep

## Upstream ask

`@dfinity/oisy-wallet-signer` 6.0.0's peer regression looks unintentional (a major bump that moves *down* a major on core, and tightens two peers to `~`). Worth raising with that team — if 6.x is meant to be the core-5 line, its peers need correcting.

marc0olo:
Three skills recommended installs that silently resolved `@icp-sdk/core` to the wrong major — `wallet-integration` down to core **4**, `vetkeys`/`encrypted-maps` up to core **6** — so a dapp combining either with Internet Identity login could not produce a valid dependency tree. Every skill now names `@icp-sdk/core@^5` explicitly and pins the two dependencies whose `latest` drifts off it.

**For the reviewer:** the diff is small (9 files, +55/−21). One eval behaviour fails, and it is pre-existing and filed as #367 — details at the bottom. Everything else is green.

Closes #364.

## Problem

`skills/wallet-integration/SKILL.md` installed `@icp-sdk/core` unpinned. Run today, that resolved to **core 4.2.3** — two majors below every other IC skill — and contradicted the same skill's own Prerequisites, which stated `>= 5.0.0`:

```
add @icp-sdk/core 4.2.3        <- two majors below every other skill
add @icp-sdk/canisters 3.1.0   <- downgraded from 3.6.0
add @dfinity/utils 4.0.3       <- downgraded from 4.2.1
```

**Cause:** `@dfinity/oisy-wallet-signer` 6.0.0 regressed its peers *backwards* — `@icp-sdk/core` `^5` → `^4`, `@dfinity/utils` → `~4.0.3`, `@icp-sdk/canisters` → `~3.1.0`. The skill was written correctly against 4.1.x; the unpinned command took 6.0.0.

**Impact:** `internet-identity` (→ `@icp-sdk/auth`, peer core `^5`) and `wallet-integration` (→ core `^4`) could not be used in the same project — a common combination.

## Changes

| File | Change |
|---|---|
| `wallet-integration/SKILL.md` | Pin oisy `^4.1.3` + core `^5`; Prerequisites list the peers 4.1.3 actually declares; new **pitfall 11** on the 6.0.0 regression |
| `vetkeys/SKILL.md` | `@icp-sdk/vetkeys` → `^0.7`; name `@icp-sdk/core@^5` explicitly in the install |
| `encrypted-maps/SKILL.md` | Same pin guidance |
| `icp-cli/references/binding-generation.md` | Remove the false *"core starts at 5.x — there is no 0.x or 1.x release"* claim; correct *"bindgen depends on core"*; bindgen floor → `^0.4.0` |
| `icp-cli/references/dfx-migration.md` | Pin the install |
| `icp-cli/SKILL.md` | Pitfall 12 pins `^5` with the `ERESOLVE` reason |
| `evaluations/{icp-cli,wallet-integration,vetkeys}.json` | One new case each for wallet + vetkeys; two icp-cli cases updated |

Two factual corrections worth calling out:

1. **"`@icp-sdk/core` starts at version 5.x — there is no 0.x or 1.x release"** was false. The published line is `1.0.0-beta.0…5` → `4.0.0…4.2.3` (stable) → `5.x` → `6.x`, and it failed to warn that 6.x now exists — the sentence that would push someone onto a broken install.
2. **"bindgen depends on `@icp-sdk/core`"** was false — bindgen's only dependency is `commander`. The code it *generates* imports core.

## Verification

The whole skill surface, installed together:

```
add @icp-sdk/core 5.4.0       add @icp-sdk/auth 8.0.3       add @icp-sdk/vetkeys 0.7.0
add @icp-sdk/signer 5.6.3     add @icp-sdk/canisters 3.6.0  add @dfinity/oisy-wallet-signer 4.1.3
add @dfinity/utils 4.2.1
```

One copy of `@icp-sdk/core`, no `ERESOLVE`. `npm run validate` — 29 skills passed, 19 warnings, identical to baseline.

## Why `^5` and not `^6`

Core 6.1.0 is published, but `@icp-sdk/auth` (8.0.3), `@icp-sdk/signer` (5.6.3) and `@dfinity/oisy-wallet-signer` (4.1.3) all still require `^5`. Installing core 6 alongside them is a hard `ERESOLVE`.

`@icp-sdk/vetkeys` **0.7.0** now peers core as `^5.0.0 || ^6.0.0`, so it no longer constrains the choice in either direction — but that makes naming core explicitly *more* important, not less: `npm i @icp-sdk/vetkeys` on its own resolves core to `6.1.0`, the highest in the range. The move to `^6` is tracked in **#365** and deliberately not done here.

<details>
<summary>Note on <code>@icp-sdk/vetkeys</code>, which moved twice while this PR was open</summary>

`0.6.0` shipped mid-review depending on `@icp-sdk/core@^6.1.0` as a plain **dependency**, not a peer. Beside `@icp-sdk/auth` that produced no `ERESOLVE` — npm simply installed two copies of core. This PR briefly pinned `^0.5` to avoid it, and the packaging was reported upstream as [dfinity/vetkeys#433](https://github.com/dfinity/vetkeys/issues/433).

`0.7.0` fixed it by making core a peer spanning both majors, so the pin was lifted to `^0.7`. Verified: vetkeys 0.7 + auth resolves to a single `@icp-sdk/core@5.4.0`, fully deduped. The upgrade is a drop-in — the public typings are byte-identical between 0.5.0 and 0.7.0 (69 exported declarations either side, empty diff), so no code samples changed.

</details>

## Evals

Version literals are deliberately kept out of the expected behaviours except the `^5` pin itself, which *is* the guidance. A case asserting a point-in-time snapshot of npm goes stale on every upstream release — the vetkeys case below was rewritten three times in a day before being made behavioural. Detecting that the pin itself has gone stale is the job of the guard proposed in #366, not of an eval.

<details>
<summary>Cases added/changed — all run with baseline</summary>

**`wallet-integration` case 5 — NEW**

```
  WITH skill: 4/4 passed
    ✅ Pins @dfinity/oisy-wallet-signer below 6.0.0 (e.g. ^4.1.3)
    ✅ Pins @icp-sdk/core to ^5
    ✅ Does NOT give an unpinned install command for @dfinity/oisy-wallet-signer or @icp-sdk/core
    ✅ Notes that the latest oisy-wallet-signer (6.0.0) peers @icp-sdk/core ^4 and is
       therefore incompatible with @icp-sdk/auth

  WITHOUT skill: 0/4 passed
    ❌ Pins oisy below 6.0.0     → installs with no version specifier, pulling latest (6.0.0)
    ❌ Pins @icp-sdk/core to ^5  → does not mention @icp-sdk/core at all
    ❌ No unpinned install       → the output is exactly an unpinned install command
    ❌ Notes the incompatibility → no compatibility discussion present
```

**`vetkeys` case 6 — NEW**

```
  WITH skill: 3/3 passed
    ✅ Installs @icp-sdk/vetkeys — NOT the legacy @dfinity/vetkeys
    ✅ Names @icp-sdk/core explicitly in the install, pinned to ^5
    ✅ Explains that vetKeys accepts more than one @icp-sdk/core major, so an install that
       omits core resolves the highest one — which @icp-sdk/auth may not accept

  WITHOUT skill: 0/3 passed
    ❌ Installs @icp-sdk/vetkeys  → installs the legacy @dfinity/vetkeys instead
    ❌ Names core explicitly      → the command does not mention @icp-sdk/core at all
    ❌ Explains the peer range    → no explanation of any kind; a bare install command
```

**`icp-cli` case 6 — "Frontend TypeScript bindings"** (two behaviours updated)

```
  WITH skill: 6/6 passed          WITHOUT skill: 0/6 passed
    ✅ Recommends @icp-sdk/bindgen with an explicit version constraint, or references
       references/binding-generation.md which carries it
    ✅ Pins @icp-sdk/core to ^5 — does NOT leave the install unpinned
    (baseline instructs 'dfx generate', which does not exist in icp-cli)
```

**`icp-cli` case 15 — "Full-stack Motoko config artifacts"** (two behaviours updated)

```
  WITH skill: 5/6 passed          WITHOUT skill: 4/6 passed
    ✅ Uses 'mops generate candid backend' as the command producing the committed .did file
    ✅ Pins @icp-sdk/core to ^5 if a version is referenced
    ❌ Uses the @icp-sdk/bindgen Vite plugin with a didFile path pointing to the committed .did file
       → the plugin config uses a 'candid' key instead of the expected 'didFile' field
```

The one failure is the pre-existing #367 issue described below, not a regression. Note this case's core behaviour ("if a version is referenced") is satisfied vacuously by the baseline, so its 4/6 is not a meaningful comparison — the discriminating behaviours here are the recipe and `mops generate candid` ones.

</details>

## ⚠️ Known failure for the reviewer to acknowledge — #367

`icp-cli` case 15 does not reach 6/6. **The failing behaviour is pre-existing and not caused by this PR** — the model writes the bindgen Vite plugin with a `candid:` key instead of `didFile:`.

Verified rather than assumed, by stashing the whole change set and re-running the case against the original content:

```
BEFORE:  ❌ ... configures the path via a 'candid' key instead of the expected 'didFile' field
AFTER:   ❌ ... uses a 'candid' key instead of the required 'didFile' key
```

Identical failure, identical reason. Root cause and suggested fix are filed as **#367**: the skill uses two different keys for "path to the `.did` file" — `candid:` (icp.yaml) appears 4× in `SKILL.md`, `didFile:` (vite.config.js) appears 0× there. Deliberately out of scope to keep this PR to version coherence.

## Related

- **#365** — move to core v6 when the chain supports it. Needs `@icp-sdk/signer` ([#39](https://github.com/dfinity/icp-js-signer/pull/39), approved but conflicting) **and** `@icp-sdk/auth`; `@icp-sdk/vetkeys` is no longer part of that chain
- **#366** — proposed version-coherence guard. Both breakages fixed here arrived via *upstream releases* with zero changes to this repo, so only a scheduled check could have caught them
- **#367** — the `didFile`/`candid` confusion above
- **#156** — separate PR for the binding-generation type representations and the `{ agent }` retraction; it touches `binding-generation.md` too, so it should land after this one


marc0olo:
Thanks — the wording point is correct and is fixed in 4d4e03b. Chasing it also surfaced something larger that the review did not catch, so this PR now covers more than it did.

## On the review comments

Both comments make the same point: *"Every IC skill pins `^5`" is inaccurate because `vetkeys/SKILL.md:25` says `^5.4`.*

Agreed. `^5.4` ⊂ `^5` — both resolve inside the 5.x line, so the *effect* was what the sentence described — but it was written as a literal string in backticks, which implies the exact text. Both instances now read:

> anchors `@icp-sdk/core` to the **5.x line** (written `^5` here, `^5.4` where a later minimum is needed)

## The larger problem

The same sentence contained a **substantive** error, not just an imprecise one:

> `@icp-sdk/auth`, `@icp-sdk/signer` and `@icp-sdk/vetkeys` all still require `^5`

`@icp-sdk/vetkeys@0.6.0` has since shipped and depends on `@icp-sdk/core@^6.1.0`. That clause was false.

It matters more than a version bump, because the failure is silent:

```bash
npm i @icp-sdk/vetkeys @icp-sdk/auth      # vetkeys unpinned — what our skills said ("≥0.5")
```
```
add @icp-sdk/vetkeys 0.6.0
add @icp-sdk/core 6.1.0     <- two copies,
add @icp-sdk/core 5.4.0     <- and no ERESOLVE
add @icp-sdk/auth 8.0.3
```

vetKeys declares `@icp-sdk/core` as a **dependency**, not a peer, so npm nests it rather than refusing. Contrast the oisy case this PR already fixed, where the peer conflict fails loudly.

And it is concretely harmful rather than merely untidy — the vetKeys clients take a caller-supplied agent:

```ts
constructor(agent: HttpAgent, canisterId: string)   // HttpAgent, Principal from @icp-sdk/core
```

so an `HttpAgent` built from core 5 (auth's copy) is passed into code typed against core 6. `Principal` crossing that boundary is the classic dual-package hazard.

Both `vetkeys` and `encrypted-maps` previously said `@icp-sdk/vetkeys (≥0.5)` — which permits `0.6.0`. Both now pin `^0.5` and explain why.

## The rule this PR now establishes

`@icp-sdk/core` anchors on the **5.x line**, because `@icp-sdk/auth` is there and effectively every real dapp authenticates. Two packages' `latest` tags leave that line **in opposite directions**:

| Package | `latest` | resolves core to | pin |
|---|---|---|---|
| `@dfinity/oisy-wallet-signer` | 6.0.0 | `^4` ⬇ | `^4.1.3` |
| `@icp-sdk/vetkeys` | 0.6.0 | `^6` ⬆ | `^0.5` |

Verified — the whole skill surface now composes with exactly one core:

```
add @icp-sdk/core 5.4.0       add @icp-sdk/auth 8.0.3       add @icp-sdk/vetkeys 0.5.0
add @icp-sdk/signer 5.6.3     add @icp-sdk/canisters 3.6.0  add @dfinity/oisy-wallet-signer 4.1.3
```

## New eval

<details>
<summary><code>vetkeys</code> case 6 — "Adversarial: vetkeys 0.6.0 duplicates @icp-sdk/core"</summary>

```
  WITH skill: 4/4 passed
    ✅ Pins @icp-sdk/vetkeys to ^0.5 — does NOT install 0.6.0 or leave the version unpinned
    ✅ Explains that @icp-sdk/vetkeys 0.6.0 requires @icp-sdk/core ^6 while @icp-sdk/auth requires ^5
    ✅ Warns that the result is duplicated/mismatched @icp-sdk/core copies rather than a clean install failure
    ✅ Installs @icp-sdk/vetkeys — NOT the legacy @dfinity/vetkeys

  WITHOUT skill: 1/4 passed
    ❌ Pins @icp-sdk/vetkeys to ^0.5
       → 'npm install @icp-sdk/vetkeys' has no version specifier at all
    ❌ Explains the ^6 vs ^5 requirement  → never mentions @icp-sdk/core
    ❌ Warns about duplicated copies      → implies a clean, unproblematic install
    ✅ Installs @icp-sdk/vetkeys — NOT the legacy @dfinity/vetkeys
```

</details>

`npm run validate` — 29 skills passed, 19 warnings, unchanged from baseline.

Note for #365: this makes the core-v6 migration more urgent than it looked. The ecosystem has now split rather than merely lagged — vetKeys is on 6, everything else on 5 — and [icp-js-signer#39](https://github.com/dfinity/icp-js-signer/pull/39) (approved, conflicting only on `package-lock.json`) is still the blocker for the rest of the chain.

marc0olo:
Fixed in e8dc66e — the inconsistency was real, though the resolution isn't the one suggested.

The comment offers two equivalent-looking options ("either make the earlier install snippet `^5.4` or change this sentence to `^5`"). They aren't equivalent: **`^5.4` has no requirement behind it any more.**

- It entered the skill because `@icp-sdk/vetkeys@0.5.0` declared `"@icp-sdk/core": "^5.4.0"`.
- `0.7.0` peers `^5.0.0 || ^6.0.0`, so the floor dropped to `5.0.0`.
- Both core APIs the skill's frontend examples use — `safeGetCanisterEnv` and the `rootKey` agent option — are present in `@icp-sdk/core@5.0.0` (verified against the 5.0.0 tarball).

Taking the first option and raising the install snippet to `^5.4` would have made the document self-consistent while propagating a constraint no package asks for. Both places now say `^5`.

One knock-on: `^5.4` was the only exception in the repo, so the hedge added earlier in this PR at `binding-generation.md:89` — *"the 5.x line (written `^5` here, `^5.4` where a later minimum is needed)"*, which came from [an earlier comment on this PR](https://github.com/dfinity/icskills/pull/368#pullrequestreview-5031113654) — is now unnecessary. It reads simply *"Every IC skill pins `@icp-sdk/core@^5` for this reason."* `grep` confirms no `@icp-sdk/core@^5.4` remains anywhere under `skills/`.

Re-ran the affected eval (`vetkeys` case 6): 3/3 with the skill, 0/3 baseline, unchanged. `npm run validate`: 29 skills passed, 19 warnings, same as baseline.

Minor note for maintainers: the review body reports "Copilot reviewed 9 out of 9 changed files", but the per-file summary table it renders is still the 6-file list from the first review pass — `skills/vetkeys/SKILL.md`, `skills/encrypted-maps/SKILL.md` and `evaluations/vetkeys.json` are missing from it, even though the new inline comment is itself on `skills/vetkeys/SKILL.md`.

marc0olo:
superseded by #400 and #401

PR #372, opened 2026-08-31
chore(upstream): sync caffeinelabs/skills 9274f9b → 38a1136

### LINKED ISSUE #5: writing-motoko: copy-paste correctness (truncated Principal.fromActor, self-contradicting break note, unseparated expressions in dot-notation example)
Copy-paste-correctness issues in the `writing-motoko` skill (we sync it verbatim downstream, so we'd rather they're fixed upstream than patched locally). Line numbers at the pinned tip.

### 1. `api-reference.md:339` — truncated `Principal.fromActor` signature
The entry is cut off mid-type:
```
- `public let fromActor : (a : actor`
```
Every other method in the list has a complete signature; this one ends at `(a : actor` with no closing paren or return type. Should be the full signature (e.g. `(a : actor {}) -> Principal`). As-is an agent reading the reference gets an invalid type and may hallucinate the rest.

### 2. `references/control-flow.md:54` — self-contradicting note on `break` (minor)
The section says:
> Motoko does not support `break` as a keyword. Use labeled loops for early exit:

…then the very next code block uses `break search;` (twice). `break` **is** a reserved keyword — it just requires a label. The guidance (labeled loops) is correct; only the sentence is off. Suggest e.g. *"Motoko has no unlabeled `break`; use a labeled loop with `break <label>`."*

### 3. `SKILL.md:83` — three expressions on one line isn't valid Motoko (minor)
In the Contextual Dot Notation block:
```motoko
// Applies to conversions too
caller.toText() myNat.toText() "hello".concat(" world") // CORRECT
Principal.toText(caller) Nat.toText(myNat) // WRONG (M0236)
```
The intent (dot-notation vs module-function form) is clear, but the `// CORRECT` line juxtaposes three separate expressions with no separators, which is a syntax error if copied — and it's inconsistent with the one-per-line, semicolon-terminated examples just above it (`map.get(key);`, `list.add(item);`). Splitting them onto separate lines (like those) would make the "CORRECT" line actually compile.

---

_Edited: I originally also listed missing `mo:core` imports in `type-conversions.md`, but on review the whole file uses import-free illustrative fragments by convention, so that's a style choice rather than a defect — removed. The "complete examples but no migration file" point remains tracked in #4._

### LINKED ISSUE #8: writing-motoko: new 'Build feedback' section reintroduces Caffeine-specific framing (cf. #1/#2)
## Context

#1 and #2 established that `writing-motoko` should stay generic — general Motoko/IC guidance, with Caffeine-platform specifics isolated or dropped. Both were fixed at commit `02e5316`.

## What changed

Commit `9274f9bb5d34` adds a new "Build feedback" section under "Toolchain (mops)":

> When building through Caffeine, enable complete subprocess output with `CAFFEINE_VERBOSE=1 caffeine build`; successful subprocess warnings are otherwise hidden. Hosted Caffeine sandboxes set this environment variable globally.
>
> - **`MOPS-WASM-COMPLEXITY`**: ...
> - **`MOPS-CHECK-DEPLOY-SKIPPED`**: ...
> - **PocketIC failure**: ...
>
> ...
>
> Caffeine may repeat failed subprocess output. Deduplicate diagnostics, fix each distinct issue, and rerun the relevant check.

The `MOPS-WASM-COMPLEXITY` / `MOPS-CHECK-DEPLOY-SKIPPED` / PocketIC-failure diagnostic guidance itself is generically useful (any Motoko dev running mops v3's `--check-wasm`/`--check-deploy` could hit these). But the intro (`caffeine build`, `CAFFEINE_VERBOSE`, "Hosted Caffeine sandboxes") and closer ("Caffeine may repeat failed subprocess output") are Caffeine-platform-specific — the same kind of content #1/#2 asked to isolate or drop.

## Ask

Could the Caffeine-specific framing be dropped or isolated (same treatment as the earlier fix), keeping the generic diagnostic-interpretation guidance? Happy to have this stay open until addressed, or to hear if it's intentional this time for a different reason.

Found while syncing `writing-motoko` downstream at [dfinity/icskills#356](https://github.com/dfinity/icskills/issues/356).

### LINKED ISSUE #9: writing-motoko: new equality guidance contradicts its own examples; ages.add() regression drops a declaration
Found via Copilot review on our sync PR ([dfinity/icskills#359](https://github.com/dfinity/icskills/pull/359)), checked against `9274f9bb5d34`. Three separate points, all in body content we sync 1:1 (not something we'd patch locally):

## 1. `ages.add(...)` — `ages` is never declared (regression)

`SKILL.md` "Implicit Parameters" section:

```motoko
ages.add("Alice", 30);               // CORRECT
ages.add(Text.compare, "Alice", 30); // WRONG (M0237)
```

`ages` has no declaration anywhere above it. The **previous** version (pinned commit `6173cbc`) had one:

```motoko
let ages = Map.empty<Text, Nat>();
ages.add(Text.compare, "Alice", 30); // explicit when needed
```

The rewrite in `9274f9bb5d34` dropped the `let ages = ...` line while keeping the `ages.add(...)` calls — this is a regression, not a stylistic choice (unlike the `map`/`numbers`/`friends` placeholder-style fragments elsewhere, which were already undeclared before and are a known convention per #5). Suggest restoring a `let ages = Map.empty<Text, Nat>();` line before the snippet.

## 2. New "prefer `equal`/`compare` over `==`" guidance contradicts nearby examples

The new Contextual Dot Notation section says to use `==` only for numeric primitives with no receiver form, and prefer `.equal(...)` for everything else (Text, Principal, Bool, ...). But several examples elsewhere in the same file still use `==` on non-numeric types added or left untouched by this same commit:

- `SKILL.md` (~line 209): `users.find(func u = u.id == caller)` — `id` is `Principal`
- `SKILL.md` (~line 389): `rightTags.any(func right = left == right)` — `Text`
- `SKILL.md` (~line 624): `users.find(func u = u.name == name)` — `Text`

Not compile errors, just inconsistent with the rule the file just introduced a few sections earlier. Worth a pass to align these with `.equal(...)` (or explicitly note `==` is still fine here for some reason we're missing).

## 3. `references/equality.md` — `Point.compare` example has an undefined `Point` type

```motoko
module Point {
  public func compare(a : Point, b : Point) : Order.Order {
    switch (Nat.compare(a.x, b.x)) { ... }
  };
};
```

`Point` (with `x`/`y` fields) is never defined, and `Nat`/`Order` aren't imported. Minor — similar territory to the "illustrative fragment" convention from #5 — but since this one introduces a whole custom module or a reader could copy it more literally, a one-line `type Point = { x : Nat; y : Nat };` above it would make it self-contained.

Separately tracked: the Caffeine-specific "Build feedback" section (also flagged by the same Copilot review) is already filed as #8.

### LINKED ISSUE #371: upstream sync available — caffeinelabs/skills 9274f9b → 38a1136
## Upstream diff: `caffeinelabs/skills` `9274f9b` → `38a1136`

Commit: [`38a113689f27`](https://github.com/caffeinelabs/skills/commit/38a113689f275bb2efe5fa87e4b151c4a8899e8d)

To sync: create branch `chore/sync-upstream-skills-38a1136`, follow the
[Upstream Sync Strategy](https://github.com/dfinity/icskills/blob/main/.claude/CLAUDE.md#upstream-sync-strategy)
in CLAUDE.md, run `npm run validate`, and open a PR that closes this issue.

**Before applying:** check `.claude/upstream.md` for icskills-owned sections.
Do NOT overwrite those sections from upstream. Also check whether any owned
section is now covered by the upstream changes — if so, drop the icskills copy
and remove it from the owned list to avoid duplicating content.

#### `SKILL.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
6c6
< version: 0.1.8
---
> version: 0.2.0
12c12
<     core: ">=2.5.0"
---
>     core: ">=2.6.0"
38c38
< - `mo:core` library version 2.5.0+ (compiler `moc` 1.11.2+)
---
> - `mo:core` library version 2.6.0+ (compiler `moc` 1.11.2+)
44c44
< - **API reference for uncertain APIs**: Use [api-reference.md](api-reference.md) to verify exact method signatures when you are about to use an unfamiliar `mo:core` API or when a compile diagnostic points at an API mismatch. Do NOT guess API shapes — a targeted lookup of a symbol you are unsure about is always worth the step; skipping it to save steps ships hallucinated APIs and costs far more in compile repair.
---
> - **API reference for uncertain APIs**: Use [api-reference.md](api-reference.md) to verify exact method signatures when you are about to use an unfamiliar `mo:core` API or when a compile diagnostic points at an API mismatch. It lists only non-deprecated APIs — a symbol that is not there should not be written. Do NOT guess API shapes — a targeted lookup of a symbol you are unsure about is always worth the step; skipping it to save steps ships hallucinated APIs and costs far more in compile repair.
73,87d72
< ### Build feedback
< 
< When building through Caffeine, enable complete subprocess output with `CAFFEINE_VERBOSE=1 caffeine build`; successful subprocess warnings are otherwise hidden. Hosted Caffeine sandboxes set this environment variable globally.
< 
< - **`MOPS-WASM-COMPLEXITY`**: Focus on the canister, function, severity, limit usage, primary contributors, result, and suggested correction. Refactor the implicated function according to its largest contributors. This check is advisory; PocketIC is authoritative.
< - **`MOPS-CHECK-DEPLOY-SKIPPED`**: Focus on the canister and compatibility diagnostic. The canister was not tested, so NEVER treat this as deployment success.
< - **PocketIC failure**: Focus on the canister, descriptive message, semantic error code, measured values, and suggested correction.
<   - For `CanisterInvalidWasm`, use the descriptive message to identify the exact validation failure.
<   - For `CanisterWasmMemoryLimitExceeded`, compare peak usage with the configured limit and reduce or relocate the implicated allocations.
<   - For traps or Candid failures, inspect the trap details or generated constructor interface.
< 
< **RULE:** For unfamiliar deployment failures, follow the documentation URL in the diagnostic before changing code. Do not guess when the descriptive message is unclear.
< 
< Caffeine may repeat failed subprocess output. Deduplicate diagnostics, fix each distinct issue, and rerun the relevant check.
< 
118c103
< **Prefer `equal` / `compare` over `==`.** `==` is compiler-generated structural equality and exists only for **shared** types, so one `var` field takes a record out of shared and `==` stops compiling (M0060). Use `==` only for the numeric primitives that have no receiver form: `Nat`, `Int`, `Float`, and the sized int types declare `equal(x, y)` without a `self` parameter, so `myNat.equal(other)` fails with M0070 and `a == b` is the right call. Other receiver methods on those types (`myNat.toText()`) are fine.
---
> **`equal` / `compare` vs `==`.** Collections take `equal` and `compare` as implicit arguments, so those are the functions to write for your own records and variants. `==` is compiler-generated structural equality and exists only for **shared** types — one `var` field takes a record out of shared and `==` stops compiling (M0060) — so do not build record comparisons on it. Comparing primitives and shared fields directly with `==` is fine, and on `Nat`, `Int`, `Float`, and the sized int types it is the only form: those declare `equal(x, y)` without a `self` parameter, so `myNat.equal(other)` fails with M0070. Other receiver methods on those types (`myNat.toText()`) are fine.
264a250
> let ages = Map.empty<Text, Nat>();   // Text.compare resolved at add, from the imported Text
271a258
> type Point = { x : Nat; y : Nat };
390c377
< - Arrays have NO `.contains(...)`. Test membership with `.indexOf(element) != null` or a predicate:
---
> - Arrays DO have `.contains(element)` (`equal` is implicit, so pass only the element). Reach for `.indexOf(element)` when you need the position — it returns `?Nat`, so keep the option and use it; and for `.any(pred)` when membership is decided by a predicate rather than equality:
509c496
< - `[T]` arrays have no `contains` at all — use `.indexOf(element) != null` or `.any(func x = x == element)` for membership.
---
> - Both `List.List<T>` and `[T]` have `contains`. Use `.any(func x = ...)` when the test is a predicate, not equality.
641a629,685
> ## Error Handling: `Result`
> 
> Use `mo:core/Result` to return a failure a caller can act on. `Result<Ok, Err>` is `{ #ok : Ok; #err : Err }`, so it is a shared type and crosses the API boundary as Candid — no wrapper needed.
> 
> **Pick the return type by what the failure means:**
> 
> | The call can fail because…                          | Return                 |
> | --------------------------------------------------- | ---------------------- |
> | the caller did something the caller can fix         | `Result<Ok, Err>`      |
> | the thing simply is not there, and that is normal   | `?T`                   |
> | an invariant this code is responsible for is broken | trap (`Runtime.trap`)  |
> 
> ```motoko
> import Result "mo:core/Result";
> 
> public type BookingError = {
>   #slotTaken : { until : Time.Time };
>   #notAuthorized;
>   #unknownRoom : Nat;
> };
> 
> public shared ({ caller }) func book(roomId : Nat, at : Time.Time) : async Result.Result<Booking, BookingError> {
>   // ...
> };
> ```
> 
> **Never launder an error into `Text`.** `Result<Booking, Text>` forces every caller — including the frontend — to string-match to tell "slot taken" from "not authorized". Make `Err` a variant; put the data each failure needs inside its own tag. A `Text` payload is fine *inside* a tag when it is a message for a human, not a discriminator.
> 
> **Do not trap on caller error.** A trap rolls back the whole message and reaches the frontend as an opaque reject — the caller cannot branch on it and the user gets no actionable message. Reserve traps for "this cannot happen" (see `?? Runtime.trap(...)` above).
> 
> **Chain, do not nest.** `mapOk`, `mapErr`, and `chain` take `self`, so they are dot notation like every other self-parameter API and a pipeline stays flat. `Result.fromOption` has no `self` — it is the module-call bridge from `?T` at the edge where absence becomes a caller-visible error.
> 
> ```motoko project=result-chain filepath=src/backend/main.mo
> import Result "mo:core/Result";
> import Map "mo:core/Map";
> import Nat "mo:core/Nat";
> 
> actor {
>   type Room = { id : Nat; name : Text };
>   type RoomView = { name : Text };
>   type BookingError = { #unknownRoom : Nat; #notAuthorized };
> 
>   let rooms : Map.Map<Nat, Room>;
> 
>   func toView(room : Room) : RoomView = { name = room.name };
>   func reserve(room : Room) : Result.Result<Room, BookingError> = #ok(room);
> 
>   public query func book(roomId : Nat) : async Result.Result<RoomView, BookingError> {
>     Result.fromOption(rooms.get(roomId), #unknownRoom(roomId))
>       .chain(func room = reserve(room))
>       .mapOk(toView);
>   };
> };
> ```
> 
> Use `switch` on `#ok` / `#err` when the arms do different work; do not write `isOk`/`isErr` followed by an unwrap — that discards the payload the type was carrying.
> 
726c770
< Treat deprecation warnings as failures. Do not write `Float.fromInt(...)` in new code; `mops check --fix` reports it as deprecated. When averaging `Nat` totals into a `Float`, import the required namespaces and use the current conversion chain:
---
> Treat deprecation warnings as failures. **Every conversion is spelled `to`, on the source value** — never a `Module.fromX` call, and never a chain through the sized numeric modules. A conversion is one receiver call:
729,731c773
< import Float "mo:core/Float";
< import Int64 "mo:core/Int64";
< import Nat64 "mo:core/Nat64";
---
> import Nat "mo:core/Nat";
733,735c775
< let numerator = Float.fromInt64(Int64.fromNat64(Nat64.fromNat(sum)));
< let denominator = Float.fromInt64(Int64.fromNat64(Nat64.fromNat(count)));
< numerator / denominator
---
> let average = sum.toFloat() / count.toFloat(); // Nat.toFloat
738c778,788
< If a conversion differs from this pattern, verify the exact `mo:core` signature before writing it. Do not guess conversion names such as `Int.fromNat` or `Int64.fromNat`.
---
> | Instead of                | write             |
> | ------------------------- | ----------------- |
> | `Float.fromInt(i)`        | `i.toFloat()`     |
> | `Float.fromInt64(i)`      | `i.toFloat()`     |
> | `Nat.fromNat64(n)`        | `n.toNat()`       |
> | `Int.fromNat(n)`          | `n.toInt()`       |
> | `Nat64.fromNat(n)`        | `n.toNat64()`     |
> | `Blob.fromArray(bytes)`   | `bytes.toBlob()`  |
> | `Iter.fromArray(a)`       | `a.values()`      |
> 
> A conversion that needs several hops is a sign the wrong function was picked: verify the exact `mo:core` signature in [api-reference.md](api-reference.md), which lists only non-deprecated APIs.
758c808
< | `field fromInt is deprecated`                          | Deprecated Float conversion  | `Float.fromInt64(Int64.fromNat64(Nat64.fromNat(n)))` |
---
> | `field fromX is deprecated`                            | Deprecated `fromX` conversion | The `toX` counterpart on the source value: `n.toFloat()`, `bytes.toBlob()` |
783c833
< | `M0096` on `contains` callback                         | Predicate passed to contains | Use `find(pred) != null`; on `[T]`, `.any(pred)` or `.indexOf(e) != null` |
---
> | `M0096` on `contains` callback                         | Predicate passed to contains | `contains` takes an element; for a predicate use `.any(pred)` for a Bool, `.find(pred)` for the element |
```

</details>

#### `api-reference.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
6a7,9
> Generated from motoko-core's API lock by `scripts/generate-motoko-api-reference.mjs`.
> Deprecated APIs are omitted — if a symbol is not here, do not write it.
> 
13a17
> - `public func contains<T>(self : [T], equal : (implicit : (T, T) -> Bool), element : T) : Bool`
26,27d29
< - `public func fromIter<T>(iter : Types.Iter<T>) : [T]`
< - `public func fromVarArray<T>(varArray : [var T]) : [T]`
49a52,99
> - `public let toBlob : (self : [Nat8]) -> Blob`
> 
> ## Blob
> 
> - `public func compare(self : Blob, other : Blob) : Order.Order`
> - `public func empty() : Blob`
> - `public func equal(self : Blob, other : Blob) : Bool`
> - `public func greater(self : Blob, other : Blob) : Bool`
> - `public func greaterOrEqual(self : Blob, other : Blob) : Bool`
> - `public func isEmpty(self : Blob) : Bool`
> - `public func less(self : Blob, other : Blob) : Bool`
> - `public func lessOrEqual(self : Blob, other : Blob) : Bool`
> - `public func notEqual(self : Blob, other : Blob) : Bool`
> - `public func size(self : Blob) : Nat`
> - `public let hash : (self : Blob) -> Types.Hash`
> - `public let toArray : (self : Blob) -> [Nat8]`
> - `public let toVarArray : (self : Blob) -> [var Nat8]`
> - `public type Blob`
> 
> ## Bool
> 
> - `public func allValues() : Iter.Iter<Bool>`
> - `public func compare(self : Bool, other : Bool) : Order.Order`
> - `public func equal(self : Bool, other : Bool) : Bool`
> - `public func logicalAnd(self : Bool, other : Bool) : Bool`
> - `public func logicalNot(self : Bool) : Bool`
> - `public func logicalOr(self : Bool, other : Bool) : Bool`
> - `public func logicalXor(self : Bool, other : Bool) : Bool`
> - `public func toText(self : Bool) : Text`
> - `public type Bool`
> 
> ## Char
> 
> - `public func compare(self : Char, other : Char) : { #less; #equal; #greater }`
> - `public func equal(self : Char, other : Char) : Bool`
> - `public func greater(self : Char, other : Char) : Bool`
> - `public func greaterOrEqual(self : Char, other : Char) : Bool`
> - `public func isAlphabetic(self : Char) : Bool`
> - `public func isDigit(self : Char) : Bool`
> - `public func less(self : Char, other : Char) : Bool`
> - `public func lessOrEqual(self : Char, other : Char) : Bool`
> - `public func notEqual(self : Char, other : Char) : Bool`
> - `public let isLower : (self : Char) -> Bool`
> - `public let isUpper : (self : Char) -> Bool`
> - `public let isWhitespace : (self : Char) -> Bool`
> - `public let toNat32 : (self : Char) -> Nat32`
> - `public let toText : (self : Char) -> Text`
> - `public type Char`
55a106,160
> ## Error
> 
> - `public func isCleanReject(self : Error) : Bool`
> - `public func isRetryPossible(self : Error) : Bool`
> - `public let code : (self : Error) -> ErrorCode`
> - `public let message : (self : Error) -> Text`
> - `public let reject : (message : Text) -> Error`
> - `public type Error`
> - `public type ErrorCode`
> 
> ## Float
> 
> - `public func add(x : Float, y : Float) : Float`
> - `public func compare(x : Float, y : Float) : Order.Order`
> - `public func div(x : Float, y : Float) : Float`
> - `public func equal(x : Float, y : Float, epsilon : Float) : Bool`
> - `public func format(self : Float, fmt : { #fix : Nat8; #exp : Nat8; #gen : Nat8; #exact }) : Text`
> - `public func greater(x : Float, y : Float) : Bool`
> - `public func greaterOrEqual(x : Float, y : Float) : Bool`
> - `public func isNaN(self : Float) : Bool`
> - `public func less(x : Float, y : Float) : Bool`
> - `public func lessOrEqual(x : Float, y : Float) : Bool`
> - `public func mul(x : Float, y : Float) : Float`
> - `public func neg(x : Float) : Float`
> - `public func notEqual(x : Float, y : Float, epsilon : Float) : Bool`
> - `public func pow(x : Float, y : Float) : Float`
> - `public func rem(x : Float, y : Float) : Float`
> - `public func sub(x : Float, y : Float) : Float`
> - `public let abs : (x : Float) -> Float`
> - `public let arccos : (x : Float) -> Float`
> - `public let arcsin : (x : Float) -> Float`
> - `public let arctan : (x : Float) -> Float`
> - `public let arctan2 : (y : Float, x : Float) -> Float`
> - `public let ceil : (x : Float) -> Float`
> - `public let copySign : (x : Float, y : Float) -> Float`
> - `public let cos : (x : Float) -> Float`
> - `public let e : Float`
> - `public let exp : (x : Float) -> Float`
> - `public let floor : (x : Float) -> Float`
> - `public let fromInt64 : (x : Int64) -> Float`
> - `public let log : (x : Float) -> Float`
> - `public let max : (x : Float, y : Float) -> Float`
> - `public let min : (x : Float, y : Float) -> Float`
> - `public let nearest : (x : Float) -> Float`
> - `public let pi : Float`
> - `public let sin : (x : Float) -> Float`
> - `public let sqrt : (x : Float) -> Float`
> - `public let tan : (x : Float) -> Float`
> - `public let toFloat32 : (self : Float) -> Prim.Types.Float32`
> - `public let toInt : (self : Float) -> Int`
> - `public let toInt64 : (self : Float) -> Int64`
> - `public let toText : (self : Float) -> Text`
> - `public let trunc : (x : Float) -> Float`
> - `public type Float`
> 
62d166
< - `public func fromNat(nat : Nat) : Int`
84,87d187
< - `public let fromInt16 : (x : Int16) -> Int`
< - `public let fromInt32 : (x : Int32) -> Int`
< - `public let fromInt64 : (x : Int64) -> Int`
< - `public let fromInt8 : (x : Int8) -> Int`
114,115d213
< - `public func fromArray<T>(array : [T]) : Iter<T>`
< - `public func fromVarArray<T>(array : [var T]) : Iter<T>`
174a273
> - `public func get<T>(self : List<T>, index : Nat) : ?T`
194a294
> - `public func retain<T>(self : List<T>, predicate : T -> Bool)`
243a344
> - `public func replace<K, V>(self : Map<K, V>, compare : (implicit : (K, K) -> Order.Order), key : K, value : V) : ?V`
247a349,350
> - `public func swap<K, V>(self : Map<K, V>, compare : (implicit : (K, K) -> Order.Order), key : K, value : V) : ?V`
> - `public func take<K, V>(self : Map<K, V>, compare : (implicit : (K, K) -> Order.Order), key : K) : ?V`
281,284d383
< - `public let fromNat16 : Nat16 -> Nat`
< - `public let fromNat32 : Nat32 -> Nat`
< - `public let fromNat64 : Nat64 -> Nat`
< - `public let fromNat8 : Nat8 -> Nat`
339,340c438,440
< - `public let fromActor : (a : actor`
< - `public let fromBlob : (self : Blob) -> Principal`
---
> - `public let fromActor : (a : actor {}) -> Principal`
> - `public let fromBlob : (blob : Blob) -> Principal`
> - `public let toActor : <A <: actor {}>(p : Principal) -> A`
377a478,497
> ## Result
> 
> - `public func assertErr(self : Result<Any, Any>)`
> - `public func assertOk(self : Result<Any, Any>)`
> - `public func chain<Ok1, Ok2, Err>( self : Result<Ok1, Err>, f : Ok1 -> Result<Ok2, Err> ) : Result<Ok2, Err>`
> - `public func compare<Ok, Err>( self : Result<Ok, Err>, other : Result<Ok, Err>, compareOk : (implicit : (compare : (Ok, Ok) -> Order.Order)), compareErr : (implicit : (compare : (Err, Err) -> Order.Order)) ) : Order.Order`
> - `public func equal<Ok, Err>( self : Result<Ok, Err>, other : Result<Ok, Err>, equalOk : (implicit : (equal : Ok, Ok) -> Bool), equalErr : (implicit : (equal : (Err, Err) -> Bool)) ) : Bool`
> - `public func flatten<Ok, Err>( self : Result<Result<Ok, Err>, Err> ) : Result<Ok, Err>`
> - `public func forErr<Ok, Err>(self : Result<Ok, Err>, f : Err -> ())`
> - `public func forOk<Ok, Err>(self : Result<Ok, Err>, f : Ok -> ())`
> - `public func fromOption<Ok, Err>(x : ?Ok, err : Err) : Result<Ok, Err>`
> - `public func fromUpper<Ok, Err>( result : { #Ok : Ok; #Err : Err } ) : Result<Ok, Err>`
> - `public func isErr(self : Result<Any, Any>) : Bool`
> - `public func isOk(self : Result<Any, Any>) : Bool`
> - `public func mapErr<Ok, Err1, Err2>( self : Result<Ok, Err1>, f : Err1 -> Err2 ) : Result<Ok, Err2>`
> - `public func mapOk<Ok1, Ok2, Err>( self : Result<Ok1, Err>, f : Ok1 -> Ok2 ) : Result<Ok2, Err>`
> - `public func toOption<Ok, Err>(self : Result<Ok, Err>) : ?Ok`
> - `public func toUpper<Ok, Err>( self : Result<Ok, Err> ) : { #Ok : Ok; #Err : Err }`
> - `public type Result<Ok, Err>`
> 
380c500,503
< - `public func trap(message : Text)`
---
> - `public func envVar<system>(name : Text) : ?Text`
> - `public func envVarNames<system>() : [Text]`
> - `public func trap(errorMessage : Text) : None`
> - `public func unreachable() : None`
391a515
> - `public func deleteAll<T>(self : Set<T>, compare : (implicit : (T, T) -> Order.Order), iter : Types.Iter<T>) : Bool`
402a527
> - `public func insertAll<T>(self : Set<T>, compare : (implicit : (T, T) -> Order.Order), iter : Types.Iter<T>) : Bool`
410a536
> - `public func retainAll<T>(self : Set<T>, compare : (implicit : (T, T) -> Order.Order), predicate : T -> Bool) : Bool`
497d622
< - `public let fromChar : (c : Char) -> Text`
510c635,643
<   // No Time.compare -->> use Int.compare to compare times.
---
> 
> > No `Time.compare` — use `Int.compare` to compare times.
> 
> ## Timer
> 
> - `public func recurringTimer<system>(duration : Time.Duration, job : () -> async ()) : TimerId`
> - `public func setTimer<system>(duration : Time.Duration, job : () -> async ()) : TimerId`
> - `public let cancelTimer : TimerId -> ()`
> - `public type TimerId`
```

</details>

#### `references/equality.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
7c7
< **Default to `equal` / `compare`. Use `==` only for the primitives that have no receiver form** — `Nat`, `Int`, `Float`, and the sized int types, where `a == b` is shorter than `Nat.equal(a, b)` and needs no import.
---
> **`==` compares two values; `equal` and `compare` are functions you can pass.** Collections need the function form, so write it for the records and variants you use as keys or elements. For a direct comparison of primitives or shared fields, `==` is fine — and on `Nat`, `Int`, `Float`, and the sized int types it is the only receiver form available, where `a == b` is shorter than `Nat.equal(a, b)` and needs no import.
17c17
< Internal state records routinely have `var` fields, so code written around `==` breaks the moment a field becomes mutable. `equal`/`compare` functions do not have that failure mode, which is why they are the default even where `==` would work today.
---
> Internal state records routinely have `var` fields, so code written around `==` breaks the moment a field becomes mutable. `equal`/`compare` functions do not have that failure mode, which is why record and variant comparisons belong in them rather than in `==`.
73a74
> type Point = { x : Nat; y : Nat };
```

</details>

#### `references/type-conversions.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
4a5,6
> **Every conversion is spelled `to`, on the source value.** Widening and narrowing are both receiver calls; `Module.fromX` is deprecated. Do not write `Nat8.fromNat64(n)` even where it still compiles — write `n.toNat8()`.
> 
21c23
< let backToNat8 = Nat8.fromNat64(nat64); // Reverse
---
> let backToNat8 = nat64.toNat8(); // Narrow — traps on overflow
25c27
< Conversion chain: `Nat8 → Nat16 → Nat32 → Nat64` (widen) or reverse with `fromNatXX` (narrow).
---
> Conversion chain: `Nat8 → Nat16 → Nat32 → Nat64` (widen) and back (narrow). Both directions are `toXX` on the value.
34c36
< let backToInt8 = Int8.fromInt64(int64); // Reverse
---
> let backToInt8 = int64.toInt8(); // Narrow — traps on overflow
38c40
< Conversion chain: `Int8 → Int16 → Int32 → Int64` (widen) or reverse with `fromIntXX` (narrow).
---
> Conversion chain: `Int8 → Int16 → Int32 → Int64` (widen) and back (narrow). Both directions are `toXX` on the value.
49,51c51,53
< // Text to Nat/Int (returns optional)
< let maybeNat = Nat.fromText("42"); // ?Nat
< let maybeInt = Int.fromText("-5"); // ?Int
---
> // Text to Nat/Int (returns optional) — receiver form on the Text
> let maybeNat = "42".toNat(); // ?Nat
> let maybeInt = "-5".toInt(); // ?Int
```

</details>

#### `SKILL.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
0a1,223
> ---
> name: reviewing-motoko
> description: >-
>   Design review for Motoko backends — invariants encoded in types, enhanced
>   migration state discipline, and multi-file structure. Load when reviewing,
>   auditing, or refactoring existing .mo files rather than writing new ones.
> version: 0.1.0
> compatibility:
>   toolchain:
>     moc: ">=1.11.2"
>     mops: "3.x"
>   mops:
>     core: ">=2.6.0"
> caffeineai-subscription: [none]
> ---
> 
> # Reviewing Motoko
> 
> This skill judges whether Motoko code is **well designed**. It does not teach the language — load `writing-motoko` for syntax, `mo:core` APIs, and mechanical rules, and `migrating-motoko-actors` before changing anything that is stable state.
> 
> Four questions decide a review, in this order — the first three because each can invalidate the next, the fourth last because it is the least consequential:
> 
> 1. **Is every invariant the code relies on encoded in a type?** Where the type system *can* rule out a bad state, leaving that to a runtime check or a comment is a defect, not a style preference.
> 2. **Is state declared with the right persistence?** Under enhanced migration, stable and transient are separate worlds. Blurring them produces state that silently stops updating.
> 3. **Does the code live in the right files?** `types.mo` / `lib/` / `mixins/` / `main.mo` is a contract about what may depend on what, not a folder convention.
> 4. **Does it read like Motoko?** Contextual dot notation, inferred implicits, no annotation the call already supplies, no classes. `mops check --fix` handles most of this; only what it cannot see is the reviewer's job.
> 
> ## Procedure
> 
> **Run the mechanical pass first** — do not spend review attention on what the compiler already reports.
> 
> ```bash
> mops check --fix
> ```
> 
> Its style auto-fixes (M0236 dot notation, M0237 redundant implicits, M0223 redundant instantiation) only fire where `[moc] args` enables `-W` for them; `writing-motoko`'s `references/project-setup.md` has the setup. Where they are off, the **Conventions** findings below are all yours to make by hand.
> 
> Then read in this order — outside in, because a misplaced file makes every type inside it suspect:
> 
> 1. `mops.toml` — is there a `[canisters.*.migrations]` chain? If not, finding **B1** below outranks everything else.
> 2. **Structure** — the file layout, then each file's imports.
> 3. `types.mo` — every type declaration. This is where most findings are.
> 4. **State** — the actor fields in `main.mo`, and what each `include` passes. See [references/state-and-persistence.md](references/state-and-persistence.md).
> 5. **The API boundary** — every `public shared` / `public query` function: its parameter types, its return type, its authorization.
> 6. **Bodies** — last, and only for the smells under **Maintainability** below.
> 
> ## Severity
> 
> Report findings with one of three levels. Do not inflate: a reviewer who calls everything a blocker gets ignored.
> 
> | Level        | Meaning                                                                                 |
> | ------------ | --------------------------------------------------------------------------------------- |
> | **Blocker**  | Can corrupt or silently lose state, or lets a caller reach an unauthorized action.       |
> | **Warning**  | An invariant the type system could enforce is left to runtime, a comment, or a convention. |
> | **Note**     | Readability, duplication, or naming. Fix while you are in the file; do not gate on it.   |
> 
> State the concrete failure, not the rule. "`status : Text` accepts `"complete"`, and the three call sites spell it `"completed"`" beats "prefer variants over strings".
> 
> ## Type encoding
> 
> The rule: **if a value has a fixed set of shapes, the type must have that set of shapes.** Full recipes, including how to carry each change through a migration, are in [references/type-encoding.md](references/type-encoding.md).
> 
> ### T1. `Text` for a closed set of values — **Warning**, **Blocker** if it gates authorization
> 
> The most common defect in generated Motoko. Every producer and consumer agrees on spelling by convention and the compiler checks nothing — no exhaustiveness, no error on a typo, no error when a new state is added and half the call sites are not updated.
> 
> ```motoko
> public type Task = { id : Nat; var status : Text };           // WRONG
> public type TaskStatus = { #open; #claimed : Principal; #done : { at : Time.Time } };
> ```
> 
> Variants are shared, so this reaches the frontend as a discriminated union too. **Flag `Text` (or `Nat`, or `Bool`) whenever the code, a comment, or a doc lists the permitted values** — `== "..."` comparisons, a `switch` on a `Text` with a `case _`, role names held as `Text`. Keep `Text` for genuinely open sets, and for strings an external system owns.
> 
> ### T2. Correlated fields — **Warning**
> 
> Two fields that are only valid together belong in one variant. A `Bool` plus an optional payload has four representable combinations and two meaningful ones.
> 
> ```motoko
> // WRONG — { hasDeadline = true; deadline = null } compiles and means nothing
> { hasDeadline : Bool; deadline : ?Time.Time }
> 
> // RIGHT
> { deadline : { #none; #due : Time.Time } }   // or simply `deadline : ?Time.Time`
> ```
> 
> If a `?T` already says it, drop the `Bool` — do not introduce a variant that only restates `?T`.
> 
> ### T3. Wrong cardinality — **Note**, **Warning** when callers loop defensively
> 
> Use the type that matches the real count. A `[T]` that is always empty-or-one makes every caller write a loop or an `if size == 1`; a `?T` tells them there is nothing to iterate.
> 
> ### T4. Errors as `Text` — **Warning**
> 
> `Result<Ok, Text>` forces every caller to string-match to tell one failure from another. Make `Err` a variant with the data each case needs. See the `Result` section of `writing-motoko` for the full rule, including when to return `?T` and when to trap instead.
> 
> ### T5. `var` fields leaking into the API — **Blocker** (it does not compile) or **Warning** (it compiles but over-shares)
> 
> A record with a `var` field is not shared, so it cannot be returned from a public function. Where a project already has the internal/view split (`Task` internal, `TaskView` shared), check that:
> 
> - no public signature mentions the internal type;
> - the view is **narrower** than the internal record, not a mechanical field-for-field copy. A view that re-exports an internal `Principal` list, an audit trail, or another user's data is an information leak that compiles.
> 
> ### T6. `case _` swallowing variants — **Warning**
> 
> A catch-all arm is what turns "adding a variant tag is a compile error everywhere it matters" into "adding a variant tag silently takes the default branch". Keep `case _` only where the arms are genuinely uniform and the set is expected to grow (e.g. mapping many tags to one label). Anywhere a new tag would need new behaviour, enumerate the tags.
> 
> ### T7. Transparent aliases sold as safety — **Note**
> 
> `type UserId = Principal` is a documentation alias: any `Principal` still type-checks where a `UserId` is expected. That is fine, and usually the right trade. Just do not treat it as an enforced distinction — and if two different id types are genuinely being confused at call sites, a one-field record (`{ userId : Principal }`) or a variant tag (`#user(Principal)`) is what actually enforces it, at the cost of changing the Candid shape.
> 
> ## State and persistence
> 
> Under enhanced migration (EM) the actor's stable fields are declared **type-only** and every initial value comes from the migration chain. That separation is the point: it is what makes stable state impossible to initialize from something transient. Details in [references/state-and-persistence.md](references/state-and-persistence.md).
> 
> ### B1. No enhanced migration — **Blocker**
> 
> No `[canisters.<name>.migrations]` chain in `mops.toml` means pre-EM persistence, where stable fields carry inline initializers — so a stable field can be initialized from a transient one:
> 
> ```motoko
> transient let defaultQuota = 100;
> let quota = defaultQuota;   // stable; the initializer runs ONCE, on first install
> ```
> 
> Change `defaultQuota` to `250`, deploy, and `quota` is still `100` — forever, no warning. Point at `migrating-motoko-actors` for the conversion; everything below assumes EM.
> 
> ### S1. Inline initializer on a stable field — **Blocker**
> 
> Under EM this is M0250 / M0014, so it will not compile — but it appears constantly in code copied from pre-EM examples and in half-finished refactors. The fix is never to add `transient`: move the initial value into the pending migration's `NewActor`.
> 
> ### S2. `transient` misused — **Warning** both ways
> 
> Ask of every actor and mixin field: *if this is lost on upgrade, is the app still correct?*
> 
> - **Missing `transient`**: caches, derived indexes, rate-limit counters, and anything rebuildable are being persisted. They then have to be migrated forever, and a stale cache survives the upgrade that was supposed to clear it.
> - **Wrongly `transient`**: durable domain data silently resets to its initializer on every upgrade. Look for `transient` on anything a user would expect to still be there after a deploy.
> 
> ### S3. `var` actor field shared with a mixin — **Blocker**
> 
> `var` parameters pass by value, so the mixin mutates a copy and the actor's field never changes. Wrap it in a record and pass the record. Also flag a fresh record literal at the `include` site (`include Api({ var n = state.n })`) — each mixin then gets its own copy. See `writing-motoko`.
> 
> ### S4. Stable state inside a `mixin` block — **Blocker**
> 
> Every bare `let`/`var` at the top of a `mixin` is implicitly stable and traps at runtime with `IC0503`. State comes in as a parameter; constants go in a module; only `transient` is allowed inline.
> 
> ### S5. Migration hygiene — **Blocker**
> 
> - A migration file importing anything other than `mo:core/...` — the chain replays forever, so a project import makes it wrong the moment that type changes. (`caffeineai-lints` catches this.)
> - More than one pending migration in a build, or an edit to a migration that predates this build.
> - A stable field in `main.mo` that no migration in the chain supplies (M0254 / M0267).
> 
> An identity migration body is a **Warning** — the change was stable-compatible, so delete the file. [Full table](references/state-and-persistence.md#migration-hygiene).
> 
> ## Structure
> 
> Layers, with dependencies pointing one way only: `main.mo` → `mixins/` → `lib/` → `types.mo`.
> 
> | File            | Holds                                                 | Must not hold                          |
> | --------------- | ----------------------------------------------------- | -------------------------------------- |
> | `types.mo`      | Type declarations; small pure helpers on those types  | State, endpoints, business rules       |
> | `lib/*.mo`      | Domain logic as stateless modules, state as parameters | Public endpoints, actor fields         |
> | `mixins/*.mo`   | Public endpoints: authorize, delegate, map to a view  | Domain logic, stable state             |
> | `main.mo`       | State declarations and `include`s                     | **Any public method**, any logic       |
> | `migrations/`   | The frozen chain, one module per file                 | Project imports                        |
> 
> - **A1 — Blocker/Warning:** a public method in `main.mo`. Move it to a mixin.
> - **A2 — Warning:** a monolithic file. One `.mo` holding types, state, logic, and endpoints together.
> - **A3 — Warning:** domain logic in a mixin body. If an endpoint's body is more than authorize → delegate → map, the middle belongs in `lib/`.
> - **A4 — Warning:** wrong dependency direction — `lib/` importing a mixin, or `types.mo` importing either.
> - **A5 — Note:** a mixin that is not a feature. Mixins split by *feature* (`mixins/Bookings.mo`), not by verb (`mixins/Getters.mo`).
> - **A6 — Warning:** over-broad state injection. A mixin receiving state it never touches widens what a change can break; pass only the slices it uses.
> 
> ## Maintainability
> 
> Most of this is caught in the last, bodies pass. The exception is **M1**, the authorization blocker — it is found during the API-boundary pass (step 5), not with the body smells below.
> 
> - **M1 — Blocker:** a `public shared` function that does not check `caller`. Every state change and every read of another user's data needs authorization on the backend.
> - **M2 — Warning:** `Nat` subtraction that is not provably non-negative at the operation. It traps.
> - **M3 — Warning:** a trap where the caller could have recovered. Traps roll back the message and reach the client as an opaque reject; caller-fixable failures belong in a `Result` error variant.
> - **M4 — Note:** duplicated internal→view mapping. One `toView` per type, in `lib/` or `types.mo`.
> - **M5 — Note:** magic values — bare literals for limits, quotas, or durations. Name them in a module.
> - **M6 — Note:** a comment asserting an invariant the type could carry. Every "must be non-empty" / "only set when …" comment is a type-encoding finding in disguise. Where Motoko cannot express the invariant, keep the comment but validate the value once at the boundary rather than re-checking it at every use.
> 
> ## Conventions
> 
> Least consequential of the four axes, and `mops check --fix` fixes most of it where the `-W` flags are on — so raise these only after the passes above, and only where the tooling cannot. `writing-motoko` has the rules; this is what to look for.
> 
> - **C1 — Note:** a module-function call where the function takes `self` — `List.add(list, x)`, `Principal.toText(p)`. M0236, auto-fixed.
> - **C2 — Note:** an implicit passed explicitly (`map.add(Text.compare, k, v)`), or a type instantiation inference already resolved. M0237 / M0223, auto-fixed.
> - **C3 — Warning:** type annotations on an inline `func` passed as a **call argument** — `xs.filter(func(x : Nat) : Bool { x > 1 })`. No diagnostic catches this, so it is the reviewer's job: the call already fixes both types, and the annotation duplicates them so they can drift. Write `xs.filter(func x = x > 1)`. Instantiate the *call* (`map<In, Out>`) when M0098 demands it — never the lambda. One exception: `: async ()` on an async callback is load-bearing.
> - **C4 — Blocker:** a `class`, or any object holding functions, used as actor state. Functions are not stable, so it fails with M0131 (`declared stable but has non-stable type`). Outside state a class still is not the idiom here — a module taking `self` plus dot notation gets the same call sites and stays stable.
> 
> ## Refactoring
> 
> Design fixes in Motoko are not free: **changing a type that is reachable from a stable actor field requires a migration**, and the chain is one-way. Before proposing a type change, establish whether that type is stable state.
> 
> 1. `mops check --fix` first, so the mechanical noise is gone.
> 2. **Batch the stable-state type changes** — one pending migration per build, so they land together.
> 3. **Write the migration** (`migrating-motoko-actors`). For `Text` → variant, decide explicitly what an unrecognized string becomes rather than inventing an `#unknown` tag.
> 4. **Move code before changing it** — relocating, then editing, keeps each diff reviewable.
> 5. `mops check --fix` until clean, then `mops build` once.
> 
> ## Reporting
> 
> Group findings by severity, blockers first. Per finding: `file:line`, the level, the concrete failure it permits, and the change. Suppress findings the compiler already reports.
> 
> ```text
> Blocker  src/backend/mixins/Tasks.mo:34  (M1)
>   `claimTask` never reads `caller` — any principal can claim any task.
> 
> Warning  src/backend/types.mo:12  (T1)
>   `Task.status : Text` is a closed set, so the comparison at Tasks.mo:51
>   spells it "complete" and is never true. Stable state — needs a migration.
> ```
> 
> If nothing is wrong, say so plainly. Do not manufacture findings to fill a report.
> 
> ## Additional resources
> 
> - [references/type-encoding.md](references/type-encoding.md) — recipes per invariant, and the migration each one needs
> - [references/state-and-persistence.md](references/state-and-persistence.md) — why EM matters, the stable/transient question, migration hygiene
> - `writing-motoko` — language mechanics, `mo:core` APIs, `Result`, dot notation
> - `migrating-motoko-actors` — writing the migration a type change requires
> - `troubleshooting-motoko-migrations` — compatibility diagnostics that do not match the source
```

</details>

#### `references/state-and-persistence.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
0a1,58
> # Reviewing state and persistence
> 
> Under enhanced migration (EM), every top-level `let`/`var` in an actor or mixin is **stable** unless prefixed `transient`, and stable fields are declared **type-only** — their values come from the migration chain.
> 
> ```motoko
> actor {
>   let tasks : Map.Map<Nat, Types.Task>;                    // stable; value from the chain
>   let state : { var nextId : Nat };                        // stable; mutable counter in a record
>   transient var cache = Map.empty<Text, Types.TaskView>(); // rebuilt each start
>   include TasksApi(tasks, state);
> }
> ```
> 
> ## Why EM is not optional
> 
> Pre-EM, stable fields carry inline initializers, so a stable field can be initialized from a transient one:
> 
> ```motoko
> transient let defaultQuota = 100;
> let quota = defaultQuota;      // stable
> ```
> 
> The initializer runs **once, on first install**. Change `defaultQuota` to `250`, deploy, and `quota` is still `100` — forever, no warning, in code that reads as though the two are linked. The same shape appears with `Time.now()`, a config record, a computed limit: the transient side keeps moving and the stable side is frozen at whatever the first install saw.
> 
> EM makes it unwritable. A stable field has no initializer, so the only way to set or change one is a migration — explicit, reviewed, applied exactly once.
> 
> ## The stable/transient question
> 
> For every actor and mixin field: **if this vanished on upgrade, would the app still be correct?**
> 
> - **Yes** → `transient`. Caches, derived indexes, memoized views, rate-limit windows, capability handles. Persisting them means migrating them forever and carrying stale entries across the upgrade meant to clear them.
> - **No** → stable (no keyword). Anything a user expects to survive a deploy.
> 
> Both directions are findings. Constants are the third case: a bare `let X = ...` in an actor or mixin is stable state, so a static value (literals, tuples, immutable records, function values) belongs in a **module**, and a non-static one — anything involving a call, operator, or control flow — has to be `transient let`.
> 
> ## Three ways mixin state silently decouples
> 
> `var` parameters pass **by value**; records pass by reference. Each of these compiles, and none of the writes reach the actor:
> 
> - `var nextId : Nat;` handed to a mixin — wrap it as `let state : { var nextId : Nat }` and pass the record.
> - A record literal at the `include` site: `include TasksApi({ var nextId = state.nextId })` builds a fresh record. Pass the binding.
> - A helper that reshapes the record: if the actor declares `{ var nextId : Nat }`, helpers must accept `{ var nextId : Nat }` and write `state.nextId`. Copying into `{ var counter = ... }` mutates only the copy.
> 
> Where two mixins need the same state, declare it once and pass the same binding to both.
> 
> A bare `let`/`var` at the top of a `mixin` is implicitly stable and traps at runtime with `IC0503`. It passes compilation and fails on the deployed canister, so it is always a blocker.
> 
> ## Migration hygiene
> 
> | Finding                                                        | Level   | Why                                                                     |
> | -------------------------------------------------------------- | ------- | ----------------------------------------------------------------------- |
> | An import other than `mo:core/...`                             | Blocker | The chain replays forever; a project type that later changes breaks it. Caught by `caffeineai-lints`. |
> | Two pending migrations in one build                            | Blocker | `check-limit = 1`. Fold the second into the first.                       |
> | An edit to a migration that predates this build                | Blocker | Applied migrations are tracked by module name; the edit never runs.      |
> | A stable field in `main.mo` no migration supplies              | Blocker | Surfaces as M0254 / M0267 and breaks later upgrades.                     |
> | A feature-shaped filename (`AddPriority.mo`)                   | Warning | Invites a second file per change instead of editing the pending one. Use `YYYYMMDD_HHMMSS.mo`. |
> | An identity migration body                                     | Warning | The change was stable-compatible; delete the file.                      |
> | A magic constant in the body                                   | Warning | Derive defaults from the migration input or a domain default.            |
```

</details>

#### `references/type-encoding.md`

<details><summary>Show diff (- old upstream, + new upstream)</summary>

```diff
0a1,91
> # Encoding invariants in Motoko types
> 
> Motoko has variants, options, records, and generics — no refinement types, no non-empty list, no zero-cost newtype. So the goal is not "no invalid value can exist", it is **no invalid value can be written by accident**. Variants and options buy that almost everywhere.
> 
> ## The cost
> 
> Everything reachable from a stable actor field is persisted, so changing such a type needs a migration, and the chain is one-way. Two consequences: get types right early, and batch the fixes (`check-limit = 1` allows one pending migration per build). Before proposing a type change, check whether the type is stable-reachable — if it is, the proposal includes the migration.
> 
> ## T1. Closed set held as `Text`
> 
> **Spot it:** permitted values listed in a comment, compared with `==` against literals, or `switch`ed with a `case _` fallback. Statuses, roles, kinds, tiers, categories.
> 
> `{ var status : Text }` fails three ways, none a compile error: a capitalisation mismatch, a comparison against a spelling that is never produced, and a new state falling silently into `case _`.
> 
> Encode it as a variant, and move data that only exists in one state *inside* that state — so there is no `claimedBy : ?Principal` that can be `null` while the status says claimed:
> 
> ```motoko
> public type TaskStatus = { #open; #claimed : Principal; #done : { at : Time.Time } };
> ```
> 
> **Keep `Text`** when the set is genuinely open (display names, notes, search terms) or an external system owns the string. In the second case parse it into a variant once at the boundary and never `switch` on the raw string again.
> 
> **Migration.** Map each known string to a tag. The value of writing it out is that it forces the decision about unrecognized strings; make that decision rather than adding an `#unknown` tag nobody removes.
> 
> ```motoko project=text-to-variant filepath=src/backend/migrations/20260101_000000.mo
> import Map "mo:core/Map";
> import Nat "mo:core/Nat";
> 
> module {
>   type OldTask = { id : Nat; var status : Text };
> 
>   // Payload-free: a raw Text source cannot populate a Principal or Time
>   // payload, so this step only lifts the tags. The payloaded form above
>   // is the end state, reached only when the source data carries the value.
>   type NewStatus = { #open; #claimed; #done };
>   type NewTask = { id : Nat; var status : NewStatus };
> 
>   type OldActor = { tasks : Map.Map<Nat, OldTask> };
>   type NewActor = { tasks : Map.Map<Nat, NewTask> };
> 
>   public func migration(old : OldActor) : NewActor {
>     let tasks = old.tasks.map<Nat, OldTask, NewTask>(
>       func(_id, task) {
>         {
>           id = task.id;
>           // Unrecognized strings become #open — still actionable, and no record
>           // is silently dropped. Decide this explicitly per domain.
>           var status = switch (task.status) {
>             case ("claimed") { #claimed };
>             case ("done") { #done };
>             case _ { #open };
>           };
>         };
>       }
>     );
>     { tasks };
>   };
> };
> ```
> 
> ## T2. Correlated fields
> 
> A `Bool` beside an optional payload has four representable combinations and two meaningful ones. `{ isPaid : Bool; paidAt : ?Time.Time }` becomes `{ payment : { #unpaid; #paid : Time.Time } }`.
> 
> Do not over-apply: where a plain `?T` already carries the invariant, use `?T`. The recipe is for two or more correlated fields, or when a tag needs a name an option cannot give it.
> 
> ## T3. Cardinality
> 
> Exactly one → `T`. None or one → `?T`. Any number → `[T]`. A `[T]` that is always empty-or-one makes every caller loop or check `size() == 1`.
> 
> Motoko has no non-empty array. `(T, [T])` encodes "at least one" at a real readability cost — use it only where an empty collection is a bug callers keep guarding against.
> 
> ## T4. Error types
> 
> `Result<Ok, Text>` is T1 wearing a different hat: every caller, frontend included, has to string-match to tell failures apart. Make `Err` a variant whose tags carry what a caller needs to react. A `Text` payload *inside* a tag is fine as a human-readable message, never as the discriminator.
> 
> See the `Result` section of `writing-motoko` for choosing between `Result`, `?T`, and a trap.
> 
> ## T5. Internal vs shared views
> 
> Internal records carry `var` fields and `Map`/`Set`/`List` values, so they cannot cross a public signature and projects keep a second shared type per entity.
> 
> Review the **narrowing**, not just that the split exists. A view mirroring every internal field is a leak that compiles: flag re-exported principal lists, audit trails, internal flags, and other users' data. Keep one `toView` per type — duplicated mappers drift.
> 
> ## T6. `case _` on variants
> 
> A catch-all converts "adding a tag breaks every site that needs updating" into "adding a tag silently does the wrong thing", which is most of the value of the variant. Remove it where a new tag would need new behaviour; keep it where the arms are genuinely uniform and the set is expected to grow.
> 
> ## T7. Identifier types
> 
> `type UserId = Principal` is a transparent alias — any `Principal` type-checks in its place. It documents intent and costs nothing, so do not flag it by default. Flag it only when two id types are actually being crossed at call sites; a one-field record or a variant tag enforces the distinction, at the cost of the Candid shape and a migration if the id is stable state.
```

</details>

marc0olo:
Syncs `caffeinelabs/skills` from `9274f9b` → [`38a1136`](https://github.com/caffeinelabs/skills/commit/38a113689f275bb2efe5fa87e4b151c4a8899e8d).

Closes #371

## Scope

| Skill | Upstream change | Action |
|---|---|---|
| `writing-motoko` | `0.1.8` → `0.2.0`, 4 files changed | Content synced |
| `reviewing-motoko` | none — byte-identical to our `90a2433` pin | Pin bumped, pin note dropped |
| `migrating-motoko-actors` | none | Pin bumped |
| `troubleshooting-motoko-migrations` | none | Pin bumped |

All four now share one pinned commit again, so the `reviewing-motoko` pin note about temporarily diverging pins is removed.

## `writing-motoko` content changes

- **The Caffeine-specific "Build feedback" section is gone** — resolves caffeinelabs/skills#8.
- **Arrays DO have `.contains(element)`.** This reverses the previous "arrays have NO `.contains`" guidance in two places plus the `M0096` error-table row. Confirmed against the generated reference: `public func contains<T>(self : [T], equal : (implicit : (T, T) -> Bool), element : T) : Bool`.
- **Numeric conversions are spelled `toX` on the source value.** `Float.fromInt64(Int64.fromNat64(Nat64.fromNat(sum)))` becomes `sum.toFloat()`, with a new `from` → `to` mapping table. The `field fromInt is deprecated` error row generalizes to `field fromX`.
- **New `## Error Handling: Result` section** — `Result<Ok, Err>` with a variant `Err` (never `Text`), don't trap on caller-fixable error, chain via `mapOk` / `chain` / `fromOption`.
- **Equality reframed** from "prefer `equal`/`compare` over `==`" to "`equal`/`compare` **vs** `==`", explicitly permitting `==` on primitives and shared fields — resolves caffeinelabs/skills#9 (2).
- `let ages = Map.empty<Text, Nat>()` restored and `type Point = { x : Nat; y : Nat }` added to the implicit-parameter examples — resolves caffeinelabs/skills#9 (1) and (3).
- `api-reference.md` regenerated from motoko-core's API lock (+169 lines): new `Blob`, `Bool`, `Char`, `Error`, `Float`, `Result`, `Timer` sections; deprecated `fromX` entries removed; `Principal.fromActor` signature complete — resolves caffeinelabs/skills#5 (1).
- `compatibility` bumped to `core >= 2.6.0`.

## Owned sections

All re-applied, none now covered upstream: frontmatter (our `description` / `license` / string `compatibility` / `metadata`), `references/` link paths, `docs.mops.one` → "Load the `mops-cli` skill", `## Additional Resources` → `## Additional References` plus the `mops tooling` bullet.

## A note on `"42".toNat()`

`references/type-conversions.md` now spells Text parsing as a receiver call:

```motoko
let maybeNat = "42".toNat(); // ?Nat
let maybeInt = "-5".toInt(); // ?Int
```

**This is correct.** `toNat`/`toInt` with a `Text` receiver are defined in the **target** modules — `Nat.mo:68` is `public let toNat : (self : Text) -> ?Nat = fromText;` ("meant to be used with contextual-dot notation"), and `Int.mo:118` is the equivalent for `toInt`. Our own synced `api-reference.md` lists them at line 385 (`## Nat`) and line 184 (`## Int`), since it groups by *defining* module rather than by receiver type.

An earlier revision of this description claimed the snippet was broken. That was my error — I searched `Text.mo` and the `## Text` section, found nothing, and read absence as proof. Retracted upstream in caffeinelabs/skills#10.

One genuine caveat survives, and it is a documentation gap rather than a defect: **the module that must be imported is the one that defines the function, not the receiver's.** `"42".toNat()` needs `import Nat "mo:core/Nat"`, and `bytes.toBlob()` needs `Array` (`toBlob : (self : [Nat8]) -> Blob` lives there). `--implicit-package=core` removes the requirement entirely — moc describes it as *"allow contextual dot and implicits resolution from all modules in the given package"* — and the Caffeine template sets it, but `references/project-setup.md` does not mention it. Raised upstream on #10; tracked on our side for a follow-up.

## Evaluations

Three new cases seeded from the diff — the array `contains` reversal, the `toX` conversion rule, and the new `Result` guidance are exactly where an agent hallucinates without the updated skill.

<details><summary>New cases — with skill vs baseline</summary>

|---|---|---|
| 25. Arrays have contains for membership | **4/4** | 1/4 |
| 26. Numeric conversion uses toX on the source value | **4/4** | 1/4 |
| 27. Result error type is a variant, not Text | **5/5** | 3/5 |

```
━━━ Arrays have contains for membership ━━━
  WITH skill: 4/4 passed
    ✅ Uses `tags.contains("urgent")` — `[T]` arrays DO have `contains` in mo:core
    ✅ Passes only the element — `equal` is an implicit argument
    ✅ Does NOT claim arrays have no `contains`
    ✅ Does NOT fall back to `.indexOf(...) != null` / `.any(...)` as the primary answer
  WITHOUT skill: 1/4 passed
    ❌ The primary answer uses `Array.find(tags, ...) != null` and never uses `contains`.
    ❌ The only mention of `contains` explicitly passes `Text.equal` as a third argument.
    ❌ The primary answer is a manual `Array.find(...) != null` search.

━━━ Numeric conversion uses toX on the source value ━━━
  WITH skill: 4/4 passed
    ✅ Uses receiver conversions: sum.toFloat() / count.toFloat()
    ✅ Does NOT use deprecated Module.fromX calls
    ✅ Does NOT chain through sized numeric modules (Nat64/Int64)
    ✅ Imports mo:core/Nat, not Float/Int64/Nat64
  WITHOUT skill: 1/4 passed
    ❌ Uses Float.fromInt(sum) / Float.fromInt(count) instead of toFloat().
    ❌ Explicitly uses the deprecated Float.fromInt(...).
    ❌ Imports mo:core/Float, not mo:core/Nat.

━━━ Result error type is a variant, not Text ━━━
  WITH skill: 5/5 passed
    ✅ Returns `async Result.Result<Ok, Err>` using `mo:core/Result`
    ✅ Declares the error as a variant with one tag per failure
    ✅ Does NOT use `Result<Ok, Text>` as the discriminator
    ✅ Does NOT propose trapping for caller-fixable failures
    ✅ Uses `public shared ({ caller }) func`
  WITHOUT skill: 3/5 passed
    ❌ No import from mo:core/Result is shown, so the source of Result is unconfirmed.
    ❌ Uses `public shared (msg) func`, which does not destructure the caller.
```
</details>

<details><summary>Regression check — existing cases covering rewritten content</summary>

The `contains` guidance and the equality section were both rewritten, so the two existing cases covering them were re-run:

```
━━━ contains vs find ━━━
  WITH skill: 2/2 passed

━━━ Prefer equal/compare over == on records with var fields ━━━
  WITH skill: 4/4 passed
```
</details>

`npm run validate`: 30 skills, all passed (21 warnings, unchanged from baseline).

### DIFF
--- a/.claude/upstream.md
+++ b/.claude/upstream.md
@@ -10,9 +10,9 @@ Upstream file paths and the tracking model (release-tag vs commit) are listed pe
 
 - **Upstream:** https://github.com/caffeinelabs/skills
 - **Tracking model:** commit-based (this repo has no releases/tags). Watch for changes to the skill folder between the pinned commit and `main`; the per-skill `version:` frontmatter field is a secondary signal.
-- **Commit:** 9274f9bb5d34db77c29304ea32ec5ad7bdc3d6fc
-- **Upstream version:** 0.1.8 (skill frontmatter `version:`)
-- **Last synced:** 2026-08-24
+- **Commit:** 38a113689f275bb2efe5fa87e4b151c4a8899e8d
+- **Upstream version:** 0.2.0 (skill frontmatter `version:`)
+- **Last synced:** 2026-08-31
 - **Upstream files:**
   - `skills/writing-motoko/SKILL.md`
   - `skills/writing-motoko/api-reference.md → references/api-reference.md`
@@ -23,7 +23,7 @@ Upstream file paths and the tracking model (release-tag vs commit) are listed pe
   - `skills/writing-motoko/references/reserved-keywords.md → references/reserved-keywords.md`
   - `skills/writing-motoko/references/type-conversions.md → references/type-conversions.md`
 - **icskills-owned sections (do not overwrite from upstream):**
-  - **Frontmatter (entire block):** upstream ships `version:`, an object `compatibility:` (`toolchain.moc`/`mops.core`/`mops` CLI major), and `caffeineai-subscription:`, and no `metadata:` block. We replace it with our schema: owned `description` (tuned for repo-wide trigger evals), `license: Apache-2.0`, string `compatibility` (`moc >= 1.11.2, core >= 2.5.0, mops >= 3.0.0`), and `metadata.title`/`category`.
+  - **Frontmatter (entire block):** upstream ships `version:`, an object `compatibility:` (`toolchain.moc`/`mops.core`/`mops` CLI major), and `caffeineai-subscription:`, and no `metadata:` block. We replace it with our schema: owned `description` (tuned for repo-wide trigger evals), `license: Apache-2.0`, string `compatibility` (`moc >= 1.11.2, core >= 2.6.0, mops >= 3.0.0`), and `metadata.title`/`category`.
   - **mops docs link → `mops-cli` skill:** the body's `https://docs.mops.one/` reference is rewritten to "Load the `mops-cli` skill …". Do not restore the external link on sync.
   - **`## Additional Resources` → `## Additional References`** (section renamed), plus an extra `- **mops tooling**: Load \`mops-cli\` …` bullet not in upstream.
   - **Reference-file paths:** upstream keeps `api-reference.md`/`examples.md` at the skill root; icskills places all non-SKILL files under `references/`, so intra-skill links are rewritten to `references/…`.
@@ -34,9 +34,9 @@ Upstream file paths and the tracking model (release-tag vs commit) are listed pe
 
 - **Upstream:** https://github.com/caffeinelabs/skills
 - **Tracking model:** commit-based (no releases/tags). Same as `writing-motoko`.
-- **Commit:** 9274f9bb5d34db77c29304ea32ec5ad7bdc3d6fc
+- **Commit:** 38a113689f275bb2efe5fa87e4b151c4a8899e8d
 - **Upstream version:** 0.2.2 (skill frontmatter `version:`, unchanged — no content changes in this sync)
-- **Last synced:** 2026-08-24
+- **Last synced:** 2026-08-31
 - **Upstream files:**
   - `skills/migrating-motoko-actors/SKILL.md`
   - `skills/migrating-motoko-actors/examples.md → references/examples.md`
@@ -51,9 +51,9 @@ Upstream file paths and the tracking model (release-tag vs commit) are listed pe
 
 - **Upstream:** https://github.com/caffeinelabs/skills
 - **Tracking model:** commit-based (no releases/tags). Same as `writing-motoko`.
-- **Commit:** 9274f9bb5d34db77c29304ea32ec5ad7bdc3d6fc
+- **Commit:** 38a113689f275bb2efe5fa87e4b151c4a8899e8d
 - **Upstream version:** 0.1.3 (skill frontmatter `version:`, unchanged — no content changes in this sync)
-- **Last synced:** 2026-08-24
+- **Last synced:** 2026-08-31
 - **Upstream file:** `skills/troubleshooting-motoko-migrations/SKILL.md`
 - **icskills-owned sections (do not overwrite from upstream):**
   - **Frontmatter (entire block):** same transform as the other two. Body is otherwise 1:1 with upstream (the `## Related skills` heading is kept as-is).
@@ -64,10 +64,9 @@ Upstream file paths and the tracking model (release-tag vs commit) are listed pe
 
 - **Upstream:** https://github.com/caffeinelabs/skills
 - **Tracking model:** commit-based (no releases/tags). Same as `writing-motoko`.
-- **Commit:** 90a2433578d0ea53fdbce3f51fb1dd3b8145ac04
+- **Commit:** 38a113689f275bb2efe5fa87e4b151c4a8899e8d
 - **Upstream version:** 0.1.0 (skill frontmatter `version:`)
-- **Last synced:** 2026-08-28
-- **Pin note:** added at `90a2433` while the other three `caffeinelabs/skills` entries were still pinned at `9274f9b` (the skill did not exist there). Pins realign on the next full `caffeinelabs/skills` sync — until then, a sync issue may re-show this skill's content as newly added; verify against the local file instead of re-applying.
+- **Last synced:** 2026-08-31
 - **Upstream files:**
   - `skills/reviewing-motoko/SKILL.md`
   - `skills/reviewing-motoko/references/state-and-persistence.md → references/state-and-persistence.md`
--- a/evaluations/writing-motoko.json
+++ b/evaluations/writing-motoko.json
@@ -236,6 +236,37 @@
         "Mentions `persistent actor { ... }` as the per-actor alternative if the flag can't be set",
         "Does NOT suggest adding a `stable` keyword to fix M0220"
       ]
+    },
+    {
+      "name": "Arrays have contains for membership",
+      "prompt": "In Motoko with mo:core, write a single expression that checks whether the `[Text]` array `tags` contains the exact string \"urgent\". Just the expression, no surrounding function.",
+      "expected_behaviors": [
+        "Uses `tags.contains(\"urgent\")` — `[T]` arrays DO have `contains` in mo:core",
+        "Passes only the element — `equal` is an implicit argument, so does NOT pass `Text.equal` explicitly",
+        "Does NOT claim arrays have no `contains`",
+        "Does NOT fall back to `tags.indexOf(\"urgent\") != null` or `tags.any(func t = t == \"urgent\")` as the primary answer"
+      ]
+    },
+    {
+      "name": "Numeric conversion uses toX on the source value",
+      "prompt": "In Motoko with mo:core, `sum` and `count` are both `Nat`. Write the expression computing their average as a `Float`, plus the import it needs. No surrounding function, no deploy steps.",
+      "expected_behaviors": [
+        "Uses receiver conversions on the source values: `sum.toFloat() / count.toFloat()`",
+        "Does NOT use deprecated `Module.fromX` calls such as `Float.fromInt(...)`, `Float.fromInt64(...)`, or `Nat64.fromNat(...)`",
+        "Does NOT chain through the sized numeric modules (`Nat64` / `Int64`) to reach `Float`",
+        "Imports `mo:core/Nat` (the module supplying `Nat.toFloat`), not `Float`/`Int64`/`Nat64` for the conversion"
+      ]
+    },
+    {
+      "name": "Result error type is a variant, not Text",
+      "prompt": "Write the Motoko error type and the update-method signature for `book(roomId : Nat)`, which can fail because the room is unknown, the slot is already taken, or the caller is not authorized. Just the type and the signature — no body, no deploy steps.",
+      "expected_behaviors": [
+        "Returns `async Result.Result<Ok, Err>` using `mo:core/Result`",
+        "Declares the error as a variant with one tag per failure (e.g. `#unknownRoom : Nat`, `#slotTaken`, `#notAuthorized`)",
+        "Does NOT use `Result<Ok, Text>` or any `Text` error as the discriminator",
+        "Does NOT propose trapping (`Runtime.trap`) for these caller-fixable failures",
+        "Uses `public shared ({ caller }) func` since authorization is one of the failure cases"
+      ]
     }
   ],
   "trigger_evals": {
--- a/skills/writing-motoko/SKILL.md
+++ b/skills/writing-motoko/SKILL.md
@@ -2,7 +2,7 @@
 name: writing-motoko
 description: "Motoko language pitfalls, modern syntax, and architecture patterns for the Internet Computer. Covers persistent actors, stable types, mo:core standard library, dot notation, mixins, and common compilation errors. Use when writing Motoko canister code, fixing Motoko compiler errors, or generating Motoko actors. Do NOT use for deployment, icp.yaml, or CLI commands; for design review or audit of existing Motoko code, load reviewing-motoko instead."
 license: Apache-2.0
-compatibility: "moc >= 1.11.2, core >= 2.5.0, mops >= 3.0.0"
+compatibility: "moc >= 1.11.2, core >= 2.6.0, mops >= 3.0.0"
 metadata:
   title: Writing Motoko
   category: Motoko
@@ -30,13 +30,13 @@ Motoko is an under-represented language for the Internet Computer Protocol, so y
 
 **ALWAYS use:**
 
-- `mo:core` library version 2.5.0+ (compiler `moc` 1.11.2+)
+- `mo:core` library version 2.6.0+ (compiler `moc` 1.11.2+)
 - Contextual dot notation -- `list.add(item)`, `map.get(key)`
 - Null coalesce `??` for unwrap-or-default and unwrap-or-trap (`opt ?? default`, `opt ?? Runtime.trap(...)`) -- prefer over a two-arm `switch` on `?T` (requires `moc >= 1.7.0`)
 - Plain `break` / `continue` to exit or skip a loop iteration -- they work inside `for`, `while`, and `loop` just like in other languages
 - Enhanced orthogonal persistence (state persists without `stable` keyword)
 - Principled Motoko Architecture -- `types.mo` (types), `lib/` (domain logic), `mixins/` (API endpoints), `main.mo` (composition root, NO public methods)
-- **API reference for uncertain APIs**: Use [api-reference.md](references/api-reference.md) to verify exact method signatures when you are about to use an unfamiliar `mo:core` API or when a compile diagnostic points at an API mismatch. Do NOT guess API shapes — a targeted lookup of a symbol you are unsure about is always worth the step; skipping it to save steps ships hallucinated APIs and costs far more in compile repair.
+- **API reference for uncertain APIs**: Use [api-reference.md](references/api-reference.md) to verify exact method signatures when you are about to use an unfamiliar `mo:core` API or when a compile diagnostic points at an API mismatch. It lists only non-deprecated APIs — a symbol that is not there should not be written. Do NOT guess API shapes — a targeted lookup of a symbol you are unsure about is always worth the step; skipping it to save steps ships hallucinated APIs and costs far more in compile repair.
 
 **When encountering compilation errors:** Re-check [api-reference.md](references/api-reference.md) for exact method signatures.
 
@@ -64,21 +64,6 @@ All configuration is in `mops.toml`. Load the `mops-cli` skill for `mops.toml` c
 - **`mops build`** (slow — run ONCE at the end) — Produces the compiled `.wasm` and the candid interface file `.did`. Use only as final verification after `mops check --fix` passes; never put `mops build` inside the fix loop. The `.did` file drives generated client bindings — never edit it manually.
 
 If `mops check --fix` fails: read stderr first. Do NOT call `moc` directly. Fix `.mo` source and rerun the check.
-
-### Build feedback
-
-When building through Caffeine, enable complete subprocess output with `CAFFEINE_VERBOSE=1 caffeine build`; successful subprocess warnings are otherwise hidden. Hosted Caffeine sandboxes set this environment variable globally.
-
-- **`MOPS-WASM-COMPLEXITY`**: Focus on the canister, function, severity, limit usage, primary contributors, result, and suggested correction. Refactor the implicated function according to its largest contributors. This check is advisory; PocketIC is authoritative.
-- **`MOPS-CHECK-DEPLOY-SKIPPED`**: Focus on the canister and compatibility diagnostic. The canister was not tested, so NEVER treat this as deployment success.
-- **PocketIC failure**: Focus on the canister, descriptive message, semantic error code, measured values, and suggested correction.
-  - For `CanisterInvalidWasm`, use the descriptive message to identify the exact validation failure.
-  - For `CanisterWasmMemoryLimitExceeded`, compare peak usage with the configured limit and reduce or relocate the implicated allocations.
-  - For traps or Candid failures, inspect the trap details or generated constructor interface.
-
-**RULE:** For unfamiliar deployment failures, follow the documentation URL in the diagnostic before changing code. Do not guess when the descriptive message is unclear.
-
-Caffeine may repeat failed subprocess output. Deduplicate diagnostics, fix each distinct issue, and rerun the relevant check.
 
 ## Modern Motoko Features
 
@@ -110,7 +95,7 @@ Principal.equal(a, b) // OK
 
 ```
 
-**Prefer `equal` / `compare` over `==`.** `==` is compiler-generated structural equality and exists only for **shared** types, so one `var` field takes a record out of shared and `==` stops compiling (M0060). Use `==` only for the numeric primitives that have no receiver form: `Nat`, `Int`, `Float`, and the sized int types declare `equal(x, y)` without a `self` parameter, so `myNat.equal(other)` fails with M0070 and `a == b` is the right call. Other receiver methods on those types (`myNat.toText()`) are fine.
+**`equal` / `compare` vs `==`.** Collections take `equal` and `compare` as implicit arguments, so those are the functions to write for your own records and variants. `==` is compiler-generated structural equality and exists only for **shared** types — one `var` field takes a record out of shared and `==` stops compiling (M0060) — so do not build record comparisons on it. Comparing primitives and shared fields directly with `==` is fine, and on `Nat`, `Int`, `Float`, and the sized int types it is the only form: those declare `equal(x, y)` without a `self` parameter, so `myNat.equal(other)` fails with M0070. Other receiver methods on those types (`myNat.toText()`) are fine.
 
 Your own records and variants get nothing derived — a record `compare` must be an explicit function, and custom variants need both `equal` and `compare` written out. See [references/equality.md](references/equality.md).
 
@@ -257,6 +242,7 @@ note: Did you mean to import mo:core/Int or mo:core/Nat?
 Do **not** pass the comparator explicitly when it can be inferred; that is M0237, which `mops check --fix` removes:
 
 ```motoko
+let ages = Map.empty<Text, Nat>();   // Text.compare resolved at add, from the imported Text
 ages.add("Alice", 30);               // CORRECT
 ages.add(Text.compare, "Alice", 30); // WRONG (M0237)
 ```
@@ -264,6 +250,7 @@ ages.add(Text.compare, "Alice", 30); // WRONG (M0237)
 A custom key type works the same way — give its module a `compare` and it is inferred:
 
 ```motoko
+type Point = { x : Nat; y : Nat };
 module Point {
   public func compare(a : Point, b : Point) : Order.Order { ... };
 };
@@ -382,7 +369,7 @@ Core collections (`List.List<T>`, `Map.Map<K, V>`, `Set.Set<T>`, `Queue.Queue<T>
 - After `let snapshot = list.toArray()`, only use array operations whose exact signatures are shown here or verified in the API reference — with receiver dot notation: `snapshot.filter(pred)`, `snapshot.map(mapper)`, `snapshot.sort(comparator)`, `snapshot.concat([item])`. Do not call those as module functions such as `Array.filter<T>(snapshot, pred)` or `Array.append(snapshot, [item])`.
 - If a value is an array (`[T]`) or came from `.toArray()` / `.filter(...)`, then `.map(...)` already returns an array; do not append `.toArray()` to that array-map result. (`List.List<T>.map(...)` returns a `List`, so it still needs `.toArray()` when the caller expects an array.)
 - Arrays DO support predicate search: `.find(predicate) : ?T`, `.findIndex(predicate) : ?Nat`, `.any(predicate)`, and `.all(predicate)` are all in `mo:core/Array` (see the API reference). The JS spellings `.some(...)` / `.every(...)` do not exist — use `.any` / `.all`.
-- Arrays have NO `.contains(...)`. Test membership with `.indexOf(element) != null` or a predicate:
+- Arrays DO have `.contains(element)` (`equal` is implicit, so pass only the element). Reach for `.indexOf(element)` when you need the position — it returns `?Nat`, so keep the option and use it; and for `.any(pred)` when membership is decided by a predicate rather than equality:
 
 ```motoko
 // Overlap between two tag arrays
@@ -501,7 +488,7 @@ all.map(func todo = { ... });
 
 - **`contains(element)`** -- equality check on `List`/`Set`/etc. Does NOT take a predicate.
 - **`find(predicate)`** -- predicate search on `List.List<T>` and `[T]`. Returns `?T`.
-- `[T]` arrays have no `contains` at all — use `.indexOf(element) != null` or `.any(func x = x == element)` for membership.
+- Both `List.List<T>` and `[T]` have `contains`. Use `.any(func x = ...)` when the test is a predicate, not equality.
 
 ```motoko
 numbers.contains(3); // equal inferred from the imported Nat
@@ -634,6 +621,63 @@ switch (todos.find(func todo = todo.id == targetId)) {
 };
 ```
 
+## Error Handling: `Result`
+
+Use `mo:core/Result` to return a failure a caller can act on. `Result<Ok, Err>` is `{ #ok : Ok; #err : Err }`, so it is a shared type and crosses the API boundary as Candid — no wrapper needed.
+
+**Pick the return type by what the failure means:**
+
+| The call can fail because…                          | Return                 |
+| --------------------------------------------------- | ---------------------- |
+| the caller did something the caller can fix         | `Result<Ok, Err>`      |
+| the thing simply is not there, and that is normal   | `?T`                   |
+| an invariant this code is responsible for is broken | trap (`Runtime.trap`)  |
+
+```motoko
+import Result "mo:core/Result";
+
+public type BookingError = {
+  #slotTaken : { until : Time.Time };
+  #notAuthorized;
+  #unknownRoom : Nat;
+};
+
+public shared ({ caller }) func book(roomId : Nat, at : Time.Time) : async Result.Result<Booking, BookingError> {
+  // ...
+};
+```
+
+**Never launder an error into `Text`.** `Result<Booking, Text>` forces every caller — including the frontend — to string-match to tell "slot taken" from "not authorized". Make `Err` a variant; put the data each failure needs inside its own tag. A `Text` payload is fine *inside* a tag when it is a message for a human, not a discriminator.
+
+**Do not trap on caller error.** A trap rolls back the whole message and reaches the frontend as an opaque reject — the caller cannot branch on it and the user gets no actionable message. Reserve traps for "this cannot happen" (see `?? Runtime.trap(...)` above).
+
+**Chain, do not nest.** `mapOk`, `mapErr`, and `chain` take `self`, so they are dot notation like every other self-parameter API and a pipeline stays flat. `Result.fromOption` has no `self` — it is the module-call bridge from `?T` at the edge where absence becomes a caller-visible error.
+
+```motoko project=result-chain filepath=src/backend/main.mo
+import Result "mo:core/Result";
+import Map "mo:core/Map";
+import Nat "mo:core/Nat";
+
+actor {
+  type Room = { id : Nat; name : Text };
+  type RoomView = { name : Text };
+  type BookingError = { #unknownRoom : Nat; #notAuthorized };
+
+  let rooms : Map.Map<Nat, Room>;
+
+  func toView(room : Room) : RoomView = { name = room.name };
+  func reserve(room : Room) : Result.Result<Room, BookingError> = #ok(room);
+
+  public query func book(roomId : Nat) : async Result.Result<RoomView, BookingError> {
+    Result.fromOption(rooms.get(roomId), #unknownRoom(roomId))
+      .chain(func room = reserve(room))
+      .mapOk(toView);
+  };
+};
+```
+
+Use `switch` on `#ok` / `#err` when the arms do different work; do not write `isOk`/`isErr` followed by an unwrap — that discards the payload the type was carrying.
+
 ## Common Patterns
 
 ### Module with Self Pattern
@@ -718,19 +762,25 @@ transient let cache = Map.empty<Text, User>();   // derived; rebuilt after each 
 
 ## Numeric Conversion Hygiene
 
-Treat deprecation warnings as failures. Do not write `Float.fromInt(...)` in new code; `mops check --fix` reports it as deprecated. When averaging `Nat` totals into a `Float`, import the required namespaces and use the current conversion chain:
+Treat deprecation warnings as failures. **Every conversion is spelled `to`, on the source value** — never a `Module.fromX` call, and never a chain through the sized numeric modules. A conversion is one receiver call:
 
 ```motoko
-import Float "mo:core/Float";
-import Int64 "mo:core/Int64";
-import Nat64 "mo:core/Nat64";
+import Nat "mo:core/Nat";
 
-let numerator = Float.fromInt64(Int64.fromNat64(Nat64.fromNat(sum)));
-let denominator = Float.fromInt64(Int64.fromNat64(Nat64.fromNat(count)));
-numerator / denominator
+let average = sum.toFloat() / count.toFloat(); // Nat.toFloat
 ```
 
-If a conversion differs from this pattern, verify the exact `mo:core` signature before writing it. Do not guess conversion names such as `Int.fromNat` or `Int64.fromNat`.
+| Instead of                | write             |
+| ------------------------- | ----------------- |
+| `Float.fromInt(i)`        | `i.toFloat()`     |
+| `Float.fromInt64(i)`      | `i.toFloat()`     |
+| `Nat.fromNat64(n)`        | `n.toNat()`       |
+| `Int.fromNat(n)`          | `n.toInt()`       |
+| `Nat64.fromNat(n)`        | `n.toNat64()`     |
+| `Blob.fromArray(bytes)`   | `bytes.toBlob()`  |
+| `Iter.fromArray(a)`       | `a.values()`      |
+
+A conversion that needs several hops is a sign the wrong function was picked: verify the exact `mo:core` signature in [api-reference.md](references/api-reference.md), which lists only non-deprecated APIs.
 
 ## Security and Authorization
 
@@ -750,7 +800,7 @@ Attaching cycles to an inter-canister call (`await (with cycles = ...) <call>`) 
 | `You can use the dot notation ... contains`            | Wrong Text contains shape    | `text.toLower().contains(#text term)`       |
 | `operator may trap for inferred type Nat`              | Potentially unsafe Nat math  | Avoid `Nat` subtraction; use bounds/loops   |
 | `Int cannot produce expected type Nat`                 | Int/Nat mismatch             | `.toNat()`                                  |
-| `field fromInt is deprecated`                          | Deprecated Float conversion  | `Float.fromInt64(Int64.fromNat64(Nat64.fromNat(n)))` |
+| `field fromX is deprecated`                            | Deprecated `fromX` conversion | The `toX` counterpart on the source value: `n.toFloat()`, `bytes.toBlob()` |
 | `syntax error, unexpected token '.'`                   | Missing parens               | `#text (searchTerm.toLower())`              |
 | `syntax error, unexpected token ','`                   | Missing parens in for        | `for ((key, value) in map.entries())`       |
 | `Compatibility error [M0170]`                          | Missing migration            | Load `migrating-motoko-actors`            |
@@ -775,7 +825,7 @@ Attaching cycles to an inter-canister call (`await (with cycles = ...) <call>`) 
 | `unexpected token ';'` in function call                | Semicolon after func literal | Remove `;` before `)`                       |
 | `unbound variable X`                                   | Missing import               | `import X "mo:core/X"`                      |
 | `M0098` no best choice for type param                  | Generic needs explicit types | `list.map<In, Out>(...)`                    |
-| `M0096` on `contains` callback                         | Predicate passed to contains | Use `find(pred) != null`; on `[T]`, `.any(pred)` or `.indexOf(e) != null` |
+| `M0096` on `contains` callback                         | Predicate passed to contains | `contains` takes an element; for a predicate use `.any(pred)` for a Bool, `.find(pred)` for the element |
 | `M0009` import file does not exist                     | Wrong path                   | Relative, no `.mo` extension                |
 | `M0244 variable ... is never reassigned`               | Unneeded `var` binding       | Use `let` unless reassigned with `:=`       |
 
--- a/skills/writing-motoko/references/api-reference.md
+++ b/skills/writing-motoko/references/api-reference.md
@@ -4,6 +4,9 @@ Complete API signatures from `mo:core` library.
 
 Use this as a reference when using contextual dot notation.
 
+Generated from motoko-core's API lock by `scripts/generate-motoko-api-reference.mjs`.
+Deprecated APIs are omitted — if a symbol is not here, do not write it.
+
 ## Array
 
 - `public func all<T>(self : [T], predicate : T -> Bool) : Bool`
@@ -11,6 +14,7 @@ Use this as a reference when using contextual dot notation.
 - `public func binarySearch<T>(self : [T], compare : (implicit : (T, T) -> Order.Order), element : T) : { #found : Nat; #insertionIndex : Nat }`
 - `public func compare<T>(self : [T], other : [T], compare : (implicit : (T, T) -> Order.Order)) : Order.Order`
 - `public func concat<T>(self : [T], other : [T]) : [T]`
+- `public func contains<T>(self : [T], equal : (implicit : (T, T) -> Bool), element : T) : Bool`
 - `public func empty<T>() : [T]`
 - `public func enumerate<T>(self : [T]) : Types.Iter<(Nat, T)>`
 - `public func equal<T>(self : [T], other : [T], equal : (implicit : (T, T) -> Bool)) : Bool`
@@ -23,8 +27,6 @@ Use this as a reference when using contextual dot notation.
 - `public func foldLeft<T, A>(self : [T], base : A, combine : (A, T) -> A) : A`
 - `public func foldRight<T, A>(self : [T], base : A, combine : (T, A) -> A) : A`
 - `public func forEach<T>(self : [T], f : T -> ())`
-- `public func fromIter<T>(iter : Types.Iter<T>) : [T]`
-- `public func fromVarArray<T>(varArray : [var T]) : [T]`
 - `public func indexOf<T>(self : [T], equal : (implicit : (T, T) -> Bool), element : T) : ?Nat`
 - `public func isEmpty<T>(self : [T]) : Bool`
 - `public func isSorted<T>(self : [T], compare : (implicit : (T, T) -> Order.Order)) : Bool`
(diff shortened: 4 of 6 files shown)


marc0olo:
Checked all four Copilot comments against motoko-core and the skill's own rules. **Two are already filed upstream, one is a new genuine defect now filed, and one is incorrect** — its suggested fix would introduce a defect the skills classify as a Blocker.

None of these are patched in this PR: per the [sync policy](https://github.com/dfinity/icskills/blob/main/.claude/CLAUDE.md#what-icskills-changes-vs-upstream), body content is not icskills-owned, so genuine defects in synced files are filed upstream rather than patched locally.

---

### 1. `SKILL.md:626` — `Result` described as unconditionally shared — ✅ **correct, filed as [caffeinelabs/skills#11](https://github.com/caffeinelabs/skills/issues/11)**

`Result` carries no sharedness constraint: `Result.mo:46` is `public type Result<Ok, Err> = Types.Result<Ok, Err>` and `Types.mo:41` is `public type Result<T, E> = { #ok : T; #err : E }`. A variant is shared only if its payloads are.

This matters more than it might look, because the skill teaches at length that internal records carry `var` fields and cannot cross a public signature. So a reader following "it is a shared type … no wrapper needed" writes `Result<Task, BookingError>` with a `var` field in `Task` and hits a compile error the doc just told them wouldn't happen. The section's own second example already gets this right — it returns `Result<RoomView, …>` and defines `toView` to narrow it — so only the prose overstates it.

### 2. `SKILL.md:666` — "`let rooms : Map.Map<Nat, Room>;` needs an initializer" — ❌ **incorrect, not filed**

This is correct Motoko as written. Under enhanced migration, stable actor fields are declared **type-only** and every initial value comes from the migration chain. This skill states it in five places — `SKILL.md:23` ("Inline initializers on stable actor fields" is listed under NEVER), `:43`, `:721` ("declared with **types only — no initializers**"), `:746`, and `:858` (Critical Requirement 10) — and [`reviewing-motoko` S1](https://github.com/dfinity/icskills/blob/main/skills/reviewing-motoko/SKILL.md) classifies an inline initializer on a stable field as a **Blocker**, failing with M0250 / M0014.

Adding `Map.empty()` as suggested would introduce exactly that Blocker. Declining this one.

### 3. `type-conversions.md:53` — `"42".toNat()` — ✅ **right conclusion, already filed as [#10](https://github.com/caffeinelabs/skills/issues/10)**

Already documented in this PR's description as a known defect in the synced content.

One correction to the diagnosis, though: the comment says the examples "rely on `Nat`/`Int` being in scope (the functions live on those modules)", implying an import would fix it. It wouldn't. Contextual dot notation resolves on the **receiver's** type, so `"42".toNat()` looks for `toNat` in `Text` — and `Text` has no `toNat`/`toInt` at all (verified in `motoko-core/src/Text.mo` and in this commit's own regenerated `api-reference.md`, whose `## Text` section lists 41 entries and neither of these). No import makes it compile; the correct call is the `Nat.fromText(...)` the rewrite removed, which is still current and not deprecated.

### 4. `type-conversions.md:5` — narrow the `Module.fromX` deprecation claim — ✅ **correct, already filed as [#10](https://github.com/caffeinelabs/skills/issues/10)**

This matches the "Scope note on the surrounding rule" section of #10 almost exactly, including the suggested remedy of scoping the rule to the sized-numeric conversions. Nice independent confirmation.

---

### One more, found while verifying

Not raised by the review: the same `result-chain` example declares `public query func book(...)`, while the first snippet in that section declares the same function as `public shared ({ caller }) func book(...)`. Beyond the inconsistency, a `book` that calls `reserve` is a state mutation declared as a query — harmless in the snippet since `reserve` is a stub, but a reader filling in a real `reserve` gets a reservation that is silently discarded when the query completes. Filed as the second point of [#11](https://github.com/caffeinelabs/skills/issues/11).

### Upstream issue status after this PR

| Issue | Status |
|---|---|
| [#8](https://github.com/caffeinelabs/skills/issues/8) Caffeine "Build feedback" | Closed — fixed at `38a1136` |
| [#9](https://github.com/caffeinelabs/skills/issues/9) equality + `ages` regression | Closed — all 3 points fixed |
| [#5](https://github.com/caffeinelabs/skills/issues/5) copy-paste correctness | 2 of 3 fixed; asked upstream whether the last is intentional |
| [#3](https://github.com/caffeinelabs/skills/issues/3) dropped Motoko items | `M0141` outstanding; asked upstream whether intentional |
| [#10](https://github.com/caffeinelabs/skills/issues/10) `"42".toNat()` regression | Filed |
| [#11](https://github.com/caffeinelabs/skills/issues/11) `Result` sharedness + `query` mutation | Filed |

marc0olo:
## Correction to my review reply above

A `caffeinelabs/skills` maintainer pointed out that `"42".toNat()` compiles — you just need `Nat` in scope (or `--implicit-package=core`). **They're right, my analysis above was wrong, and Copilot's comment #3 was correct.**

`toNat`/`toInt` with a `Text` receiver are defined in the **target** modules, not in `Text`:

```motoko
// motoko-core/src/Nat.mo:68
/// This functions is meant to be used with contextual-dot notation.
/// assert "1234".toNat() == ?1234;
public let toNat : (self : Text) -> ?Nat = fromText;

// motoko-core/src/Int.mo:118
public func toInt(self : Text) : ?Int { fromText(self) };
```

I assumed a receiver-typed function lives in the receiver's module, searched `Text.mo` and the `## Text` section of `api-reference.md`, and read absence as proof. But the reference groups by **defining** module — `toNat` is at line 385 under `## Nat` and `toInt` at line 184 under `## Int`, in the same file I cited as evidence against them. My three "independent confirmations" were one error counted three times.

**Corrections to the four review comments, restated accurately:**

| # | Comment | Correct verdict |
|---|---|---|
| 1 | `Result` described as unconditionally shared | ✅ Valid — [#11](https://github.com/caffeinelabs/skills/issues/11) stands, unaffected by this |
| 2 | `let rooms : Map.Map<Nat, Room>;` needs an initializer | ❌ Still incorrect — type-only declaration is required under EM |
| 3 | `"42".toNat()` relies on `Nat`/`Int` being in scope | ✅ **Correct, and I was wrong to dismiss the mechanism.** This is exactly right: the functions do live on those modules, and without an import (or `--implicit-package=core`) the snippet does not compile |
| 4 | Narrow the `Module.fromX` deprecation claim | ⚠️ Weaker than I said — `Text`→`Nat`/`Int` *does* have a receiver form, which is why the rule was written that way. Some genuine `fromX` constructors remain (`Text.fromArray`, `Principal.fromText`), so it's mildly over-general, but not the defect I described |

So: **only comment 2 is actually wrong**, and my reply above got comments 3 and 4 backwards. [#10](https://github.com/caffeinelabs/skills/issues/10) is retracted and closed; the PR description is updated.

**Nothing in the synced content needs changing** — this branch is correct as-is, and the owned-patch plan discussed offline is dropped, since there's nothing to patch.

### The one real gap this surfaced

Not a defect in the sync, but worth tracking: **which module to import is not guessable from the call site.** `"42".toNat()` needs `import Nat`, `bytes.toBlob()` needs `Array` (`toBlob : (self : [Nat8]) -> Blob` is defined there). An agent whose instinct on failure is `import Text` gets nowhere.

`--implicit-package=core` removes the requirement — moc: *"allow contextual dot and implicits resolution from all modules in the given package"* — and our `caffeine-app` skill sets it in its `mops.toml`, but `writing-motoko/references/project-setup.md`, whose entire job is the one-time `[moc] args`, doesn't mention it. So a self-managed project set up from our own skill hits the import requirement that a Caffeine project never sees.

Raised upstream on #10 with an offer to PR it. Not blocking this sync.

raymondk (approved):

PR #382, opened 2026-09-11
feat(autoupgrade-icp-cli): new skill for keeping icp and ic-wasm current

raymondk:
## What

A new skill, `autoupgrade-icp-cli` — the CLI counterpart to `autosync-ic-skills`. It is a one-time installer: it registers a `SessionStart` hook plus `.claude/upgrade-icp-cli.sh`, which compares the installed `icp` and `ic-wasm` against the latest release and either reports the upgrade (**notify**, the default) or applies it (**auto**).

The installer asks the user which mode they want rather than picking one, because auto mode upgrades the build toolchain while a project is open — what some teams want and what others pin against.

`ic-wasm` is covered alongside `icp` because the official recipes (`@dfinity/motoko`, `@dfinity/rust`, `@dfinity/static-site`, `@dfinity/asset-canister`) invoke it on every build. Upgrading only one of the pair produces build failures that read as recipe bugs.

## Why not just `icp settings update-check`

The built-in check is worth leaving on, but it only speaks once the user runs an `icp` command — by which point the agent has already chosen what to write. This hook runs before the session's first prompt, so the version reaches Claude's context up front; it also covers `ic-wasm`, which the built-in check does not.

## How the script works

It resolves each binary through its symlinks and reads the install channel off the path, because the upgrade command differs per channel and guessing wrong leaves two copies of the CLI on `PATH` shadowing each other:

| Binary resolves to | Channel | Upgrade command | Latest version from |
|---|---|---|---|
| `…/node_modules/@icp-sdk/…` | npm | `npm install -g <pkg>@latest` | GitHub release tag |
| `…/Cellar/…` | Homebrew | `brew upgrade <formula>` | `formulae.brew.sh` API |
| cargo-dist receipt in `~/.config/<app>/` | shell installer | `icp-cli-update` / `ic-wasm-update` | GitHub release tag |
| anything else | unknown | *(none — reports the release URL)* | GitHub release tag |

- **curl and POSIX tools only** — no `jq` (unlike the autosync script). The GitHub probe follows the `/releases/latest` redirect rather than calling the API, so it needs no token and cannot hit the 60-requests-per-hour anonymous limit that a per-session hook would otherwise reach. Homebrew is asked for its own formula version because the formula can trail a release by a day or two.
- **Throttled** — a check that reached a release feed is stamped in `.claude/.icp-upgrade-check` and not repeated for 6h (`ICP_UPGRADE_CHECK_INTERVAL`, `--force`, and any run from a terminal override it). Measured: ~1.1 s for a real check, ~10 ms throttled. A failed probe is deliberately not stamped, so an offline session retries at the next one.
- **Safe by construction** — never installs an absent tool (even in auto mode), never runs `sudo` (a non-writable npm prefix is reported instead, since a password prompt inside a hook has no terminal), detects when an upgrade succeeded but `PATH` still resolves to the old copy, and exits 0 on every failure path.
- For auto mode the skill registers the hook with `"timeout": 300`, since installing a CLI binary can exceed the 60 s hook default and a timed-out install leaves the tool half-written.

Verified against real installs on this machine: `icp` 1.2.0 from the shell installer (shell channel, `icp-cli-update`) and `ic-wasm` 0.9.11 from npm (npm channel) — two different channels for the two tools, each detected and given the right command. The offline, up-to-date, missing-tool, unknown-channel, non-writable-prefix, and shadowed-install paths were each exercised with stubs; `version_lt` has 9 cases covering pre-releases and multi-digit components.

## Evals

`evaluations/autoupgrade-icp-cli.json` — 5 output evals, 18 trigger queries.

**Output: 19/20 with skill vs 9/20 baseline. Triggers: 18/18.**

Eval 2 initially scored 3/4 — the agent produced a correct auto-mode hook but never mentioned that it changes the build toolchain mid-project. That risk was only stated in Step 1 and in a later section, not where the hook JSON is written; moving it next to the auto-mode block fixed it (re-run below: 4/4).

<details>
<summary>Full eval output</summary>

```
Evaluating skill: autoupgrade-icp-cli
Output cases: Setup plan and settings.json entry, Auto mode needs a raised hook timeout, Channel-correct upgrade command, ic-wasm version skew after upgrading icp, Upgrade ran but the version did not move

━━━ Setup plan and settings.json entry ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 6/6 passed
    ✅ Asks the user whether the hook should only notify or upgrade automatically
    ✅ Registers a SessionStart hook whose command runs the script
    ✅ Says the script is downloaded from the specified URL rather than hand-written
    ✅ Says an existing settings.json must be merged and not duplicated/overwritten
    ✅ Mentions the user will be asked to approve/trust the new hook
    ✅ Does not invent a settings.json schema — nests under hooks.SessionStart with hooks array of {type, command}

  WITHOUT skill: 2/6 passed
    ❌ Asks whether hook should only notify or upgrade automatically
       → The output unilaterally designs a notify-only warning script without presenting or asking about an auto-upgrade alternative.
    ✅ Registers a SessionStart hook whose command runs the script
    ❌ Says the script is downloaded from the skills.internetcomputer.org URL rather than hand-written
       → The output explicitly instructs writing a custom hand-written shell script and never mentions downloading from that or any URL.
    ❌ Says existing settings.json must be merged, preserving other keys/hooks, not duplicated/overwritten
       → The output shows a standalone JSON entry with no mention of merging with or preserving an existing settings.json file.
    ❌ Mentions user will be asked to approve/trust the new hook before it runs automatically
       → There is no mention anywhere of a trust/approval prompt for the new hook.
    ✅ Does not invent a settings.json schema

━━━ Auto mode needs a raised hook timeout ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 3/4 passed
    ✅ Uses the auto mode flag on the command (e.g. `--mode auto`)
    ✅ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → The caveats cover silent execution, no prompt, sudo avoidance, shadowed PATH installs, pre-release handling, and swallowed network errors, but never mention that auto-upgrading icp/ic-wasm mid-session could change build/toolchain behavior for the current project.
    ✅ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs

  WITHOUT skill: 1/4 passed
    ❌ Uses the auto mode flag on the command (e.g. `--mode auto`)
       → The assistant produced no hook JSON at all, only a request for file access.
    ❌ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
       → No JSON or timeout value was ever presented.
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → The output contains no discussion of risks since no solution was produced.
    ✅ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs

━━━ Channel-correct upgrade command ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 3/3 passed
    ✅ Gives `icp-cli-update` as the upgrade command
    ✅ Does NOT tell the user to run `npm install -g @icp-sdk/icp-cli@latest`
    ✅ Explains that the upgrade command depends on the channel the CLI was installed from

  WITHOUT skill: 2/3 passed
    ❌ Gives `icp-cli-update` (the updater the shell installer places on PATH) as the upgrade command
       → The output recommends re-running the curl install script instead of the `icp-cli-update` binary.
    ✅ Does NOT tell the user to run `npm install -g @icp-sdk/icp-cli@latest`, which would install a second copy from a different channel
    ✅ Explains that the upgrade command depends on the channel the CLI was installed from

━━━ ic-wasm version skew after upgrading icp ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 3/3 passed
    ✅ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
    ✅ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

  WITHOUT skill: 1/3 passed
    ❌ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
       → The output vaguely gestures at stale cached moc/ic-wasm binaries causing version skew, but never clearly identifies the specific mechanism that ic-wasm is a separately-versioned tool that doesn't get updated when icp is upgraded via npm.
    ❌ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
       → The output never suggests running `ic-wasm --version` or gives any explicit upgrade command for ic-wasm; it only suggests generically clearing caches and rebuilding.
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

━━━ Upgrade ran but the version did not move ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 4/4 passed
    ✅ Identifies that a second copy of icp from a different install channel is earlier on PATH and shadowing the npm one (commonly ~/.cargo/bin/icp from the shell installer)
    ✅ Suggests resolving the actual binary being used, e.g. `command -v icp` / `which icp` (and following the symlink) to see which copy wins
    ✅ Says upgrading through one channel will never move a copy installed through another — remove one, or upgrade the one that is actually on PATH
    ✅ Does NOT blame the npm cache or recommend reinstalling npm as the primary explanation

  WITHOUT skill: 3/4 passed
    ✅ Identifies that a second copy of icp from a different install channel is earlier on PATH and shadowing the npm one (commonly ~/.cargo/bin/icp from the shell installer)
    ✅ Suggests resolving the actual binary being used, e.g. `command -v icp` / `which icp` (and following the symlink) to see which copy wins
    ❌ Says upgrading through one channel will never move a copy installed through another — remove one, or upgrade the one that is actually on PATH
       → The output offers PATH reordering, removing the cargo copy, or aliasing as fixes, but never states the underlying principle that an npm upgrade can never touch a cargo-installed binary.
    ✅ Does NOT blame the npm cache or recommend reinstalling npm as the primary explanation

━━━ Trigger Evals ━━━

  Running trigger evaluation...

  Should trigger: 9/9 correct
    ✅ "Can you make this project always use the latest icp CLI?"
    ✅ "Set up a hook so I stop working with an outdated icp version"
    ✅ "I keep forgetting to update icp-cli. Automate it for me."
    ✅ "How do I upgrade the ICP CLI to the newest release?"
    ✅ "My icp is on 1.2.0 and the latest is 1.5.0 — what's the upgrade command?"
    ✅ "Install something that keeps icp and ic-wasm current at the start of every session"
    ✅ "I ran npm install -g @icp-sdk/icp-cli@latest but icp --version hasn't changed"
    ✅ "Is my ic-wasm too old for the version of icp I'm running?"
    ✅ "Add automatic ICP toolchain updates to .claude/settings.json"

  Should NOT trigger: 9/9 correct
    ✅ "Upgrade my canister without wiping its state — I think I need icp deploy --mode upgrade"
    ✅ "Bump the @dfinity/motoko recipe in my icp.yaml to the latest version"
    ✅ "Keep my Internet Computer skills in .claude/skills up to date automatically"
    ✅ "Update the moc compiler version in my mops.toml toolchain section"
    ✅ "My icp deploy is failing with a build error in the rust recipe"
    ✅ "What's the latest version of @icp-sdk/core I should depend on in my frontend?"
    ✅ "Write a GitHub Actions workflow that deploys my canister on every push to main"
    ✅ "How do I upgrade a canister's Wasm module on mainnet without losing stable memory?"
    ✅ "Add a SessionStart hook that runs my test suite before I start working"

━━━ Summary ━━━

  Output evals:
    Setup plan and settings.json entry: WITH 6/6 | WITHOUT 2/6
    Auto mode needs a raised hook timeout: WITH 3/4 | WITHOUT 1/4
    Channel-correct upgrade command: WITH 3/3 | WITHOUT 2/3
    ic-wasm version skew after upgrading icp: WITH 3/3 | WITHOUT 1/3
    Upgrade ran but the version did not move: WITH 4/4 | WITHOUT 3/4
  Trigger evals: should-trigger 9/9 | should-not-trigger 9/9

Full results saved to: /home/raymond/dev/icskills/evaluations/results/autoupgrade-icp-cli-2026-09-11T15-31-18-804Z.json
```

</details>

<details>
<summary>Eval 2 re-run after the fix</summary>

```
Evaluating skill: autoupgrade-icp-cli
Output cases: Auto mode needs a raised hook timeout

━━━ Auto mode needs a raised hook timeout ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 4/4 passed
    ✅ Uses the auto mode flag on the command (e.g. `--mode auto`)
    ✅ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
    ✅ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
    ✅ Does NOT claim the upgrade runs in the background

  WITHOUT skill: 1/4 passed
    ❌ Uses the auto mode flag on the command (e.g. `--mode auto`)
       → The assistant never produced any hook command or JSON at all, only asked for file access permission.
    ❌ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
       → No hook JSON or timeout value was output; the assistant stopped at a permission request.
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → No risk explanation was given since the task was not completed.
    ✅ Does NOT claim the upgrade runs in the background

━━━ Summary ━━━

  Output evals:
    Auto mode needs a raised hook timeout: WITH 4/4 | WITHOUT 1/4

Full results saved to: /home/raymond/dev/icskills/evaluations/results/autoupgrade-icp-cli-2026-09-11T15-33-26-242Z.json
```

</details>

## Checks

- `npm run validate` — passes (0 errors; the 3 repo-wide warnings are pre-existing missing-eval files for other skills)
- `npm run build` — passes; the skill and `scripts/upgrade-icp-cli.sh` are published under `.well-known/skills/autoupgrade-icp-cli/`, so the URL the skill tells agents to curl resolves once this is deployed
- `skill-validator score evaluate --provider claude-cli` — **4.50/5** (clarity 4, actionability 5, token efficiency 4, scope discipline 5, directive precision 5, novelty 4)

### DIFF
--- a/evaluations/autoupgrade-icp-cli.json
+++ b/evaluations/autoupgrade-icp-cli.json
@@ -1,0 +1,81 @@
+{
+  "skill": "autoupgrade-icp-cli",
+  "description": "Evaluation cases for the autoupgrade-icp-cli skill. Tests whether agents set the hook up correctly (ask for a mode, fetch the published script, merge settings.json, raise the timeout for auto mode) and whether they diagnose toolchain-version problems the way the skill describes: channel-correct upgrade commands, ic-wasm skew, and a shadowed second copy on PATH.",
+  "output_evals": [
+    {
+      "name": "Setup plan and settings.json entry",
+      "prompt": "I want every Claude Code session in this project to check whether my icp CLI is out of date. Describe the exact steps you'd take and show the .claude/settings.json entry you'd add. Don't run anything or edit files.",
+      "expected_behaviors": [
+        "Asks the user (or explicitly offers the choice) whether the hook should only notify or should upgrade automatically, rather than silently picking one",
+        "Registers a SessionStart hook whose command runs the script, e.g. `bash .claude/upgrade-icp-cli.sh`",
+        "Says the script is downloaded from https://skills.internetcomputer.org/.well-known/skills/autoupgrade-icp-cli/scripts/upgrade-icp-cli.sh rather than hand-written",
+        "Says an existing .claude/settings.json must be merged (preserving other keys/hooks) and the entry not duplicated, never overwritten",
+        "Mentions the user will be asked to approve/trust the new hook before it runs automatically",
+        "Does NOT invent a settings.json schema — the entry nests under hooks.SessionStart with a hooks array of {type: command, command: ...}"
+      ]
+    },
+    {
+      "name": "Auto mode needs a raised hook timeout",
+      "prompt": "Set my SessionStart hook to actually install the new icp version automatically instead of just telling me about it. Just show me the final hook JSON and note anything I should watch out for.",
+      "expected_behaviors": [
+        "Uses the auto mode flag on the command (e.g. `--mode auto`)",
+        "Raises the hook timeout above the 60 second default (e.g. \"timeout\": 300), because downloading and installing a CLI binary can exceed it",
+        "Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project",
+        "Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs"
+      ]
+    },
+    {
+      "name": "Channel-correct upgrade command",
+      "prompt": "I installed icp a while ago with the curl shell installer from the icp-cli releases page, not with npm. icp --version says 1.2.0 and there's a newer one. What's the single command to upgrade it? Just the command and one line of why.",
+      "expected_behaviors": [
+        "Gives `icp-cli-update` (the updater the shell installer places on PATH) as the upgrade command",
+        "Does NOT tell the user to run `npm install -g @icp-sdk/icp-cli@latest`, which would install a second copy from a different channel",
+        "Explains that the upgrade command depends on the channel the CLI was installed from"
+      ]
+    },
+    {
+      "name": "ic-wasm version skew after upgrading icp",
+      "prompt": "I just upgraded icp to the latest version with npm. Now `icp build` fails inside the @dfinity/motoko recipe with a metadata/optimization error, and nothing in my project changed. What should I check first? Keep it brief.",
+      "expected_behaviors": [
+        "Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build",
+        "Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)",
+        "Does NOT send the user straight into debugging their Motoko code or recipe config as the first step"
+      ]
+    },
+    {
+      "name": "Upgrade ran but the version did not move",
+      "prompt": "I ran `npm install -g @icp-sdk/icp-cli@latest`, it finished with no errors, but `icp --version` still prints the old version. What's going on? Just the diagnosis and how to confirm it.",
+      "expected_behaviors": [
+        "Identifies that a second copy of icp from a different install channel is earlier on PATH and shadowing the npm one (commonly ~/.cargo/bin/icp from the shell installer)",
+        "Suggests resolving the actual binary being used, e.g. `command -v icp` / `which icp` (and following the symlink) to see which copy wins",
+        "Says upgrading through one channel will never move a copy installed through another — remove one, or upgrade the one that is actually on PATH",
+        "Does NOT blame the npm cache or recommend reinstalling npm as the primary explanation"
+      ]
+    }
+  ],
+  "trigger_evals": {
+    "description": "Queries to test whether the skill activates correctly. 'should_trigger' queries should cause the skill to load; 'should_not_trigger' queries should NOT activate this skill.",
+    "should_trigger": [
+      "Can you make this project always use the latest icp CLI?",
+      "Set up a hook so I stop working with an outdated icp version",
+      "I keep forgetting to update icp-cli. Automate it for me.",
+      "How do I upgrade the ICP CLI to the newest release?",
+      "My icp is on 1.2.0 and the latest is 1.5.0 — what's the upgrade command?",
+      "Install something that keeps icp and ic-wasm current at the start of every session",
+      "I ran npm install -g @icp-sdk/icp-cli@latest but icp --version hasn't changed",
+      "Is my ic-wasm too old for the version of icp I'm running?",
+      "Add automatic ICP toolchain updates to .claude/settings.json"
+    ],
+    "should_not_trigger": [
+      "Upgrade my canister without wiping its state — I think I need icp deploy --mode upgrade",
+      "Bump the @dfinity/motoko recipe in my icp.yaml to the latest version",
+      "Keep my Internet Computer skills in .claude/skills up to date automatically",
+      "Update the moc compiler version in my mops.toml toolchain section",
+      "My icp deploy is failing with a build error in the rust recipe",
+      "What's the latest version of @icp-sdk/core I should depend on in my frontend?",
+      "Write a GitHub Actions workflow that deploys my canister on every push to main",
+      "How do I upgrade a canister's Wasm module on mainnet without losing stable memory?",
+      "Add a SessionStart hook that runs my test suite before I start working"
+    ]
+  }
+}
--- a/skills/autoupgrade-icp-cli/SKILL.md
+++ b/skills/autoupgrade-icp-cli/SKILL.md
@@ -1,0 +1,248 @@
+---
+name: autoupgrade-icp-cli
+description: One-time installer that keeps a project's Internet Computer CLI toolchain current. Sets up a SessionStart hook plus a script that compares the installed `icp` and `ic-wasm` against the latest release and either reports the upgrade or applies it, using whichever channel each tool was installed from (npm, Homebrew, or the shell installer). Use when a user wants to install, bootstrap, or enable automatic icp-cli / ic-wasm updates, asks to stop running an outdated `icp`, asks how to upgrade or check the version of the ICP CLI, or pastes the link to this skill. This is a one-time setup action, not ongoing IC knowledge. Do NOT use for writing icp.yaml, deploying canisters, or other icp-cli usage questions — load `icp-cli` for those.
+license: Apache-2.0
+compatibility: "curl, bash; network access to github.com (and formulae.brew.sh for Homebrew installs)"
+metadata:
+  title: Automatically Upgrade the ICP CLI
+  category: Infrastructure
+---
+
+# Set up automatic `icp` CLI upgrades
+
+This skill installs a small amount of project configuration so that **every new
+Claude Code session checks whether `icp` and `ic-wasm` are current** — and, if the
+user wants, upgrades them before any work starts.
+
+It is a **one-time installer**. After you complete the steps below, the user never
+needs this link again — the installed `SessionStart` hook does the work from then on.
+
+Both tools are covered because they ship as a pair (`npm install -g @icp-sdk/icp-cli
+@icp-sdk/ic-wasm`) and the official recipes (`@dfinity/motoko`, `@dfinity/rust`,
+`@dfinity/static-site`, `@dfinity/asset-canister`) call `ic-wasm` during every build.
+An `ic-wasm` left behind while `icp` moves forward produces build failures that look
+like recipe bugs, so upgrading only one of them is the worse default.
+
+## Why this beats the CLI's own update check
+
+`icp` already has a built-in check (`icp settings update-check`), and it is worth
+leaving enabled — but it only speaks when the user runs an `icp` command, and by then
+the agent has already chosen which commands to write. This hook runs **before** the
+session's first prompt, so:
+
+- The version lands in Claude's context up front, so it can pick flags and recipe
+  versions that exist in the installed CLI instead of discovering the mismatch from a
+  failed command.
+- `ic-wasm` is checked too, which the built-in check does not cover.
+- The upgrade command matches the channel the tool was **actually** installed from,
+  and in auto mode it is applied rather than printed.
+
+## What you will create
+
+1. `.claude/upgrade-icp-cli.sh` — the version-check script.
+2. A `SessionStart` hook in `.claude/settings.json` that runs it.
+3. An immediate first run, so the user sees the current state right away.
+
+## Step 1 — Ask the user which mode they want
+
+This is the one decision the installer cannot make for the user, so ask before writing
+anything. A CLI upgrade can change build behaviour mid-project, and some teams pin the
+toolchain deliberately:
+
+> "Two options for how the hook behaves when a newer version exists:
+> **notify** — it prints the versions and the exact upgrade command, and nothing is
+> installed until you run it. **auto** — it runs that upgrade itself at session start.
+> Which do you want? You can switch later by editing one flag in the hook."
+
+Default to **notify** if the user has no preference — it is the reversible choice, and
+switching to auto later is a one-word edit.
+
+Also tell the user what adding a hook means:
+
+> "I'm adding a `SessionStart` hook that runs `.claude/upgrade-icp-cli.sh`. Claude Code
+> will ask you to approve/trust it before it runs automatically."
+
+Do **not** attempt to bypass that approval.
+
+## Step 2 — Check prerequisites
+
+The script needs `curl` (virtually always present) and at least one of the two tools
+installed — it reports on what is there and never installs a tool the user does not
+already have:
+
+```bash
+command -v curl    >/dev/null 2>&1 && echo "curl: ok"    || echo "curl: MISSING"
+icp --version      2>/dev/null     || echo "icp: not installed"
+ic-wasm --version  2>/dev/null     || echo "ic-wasm: not installed"
+```
+
+If either tool is missing, mention the install command
+(`npm install -g @icp-sdk/icp-cli @icp-sdk/ic-wasm`) but do not run it unprompted —
+installing a toolchain is a different decision from keeping one current.
+
+Note that no `jq` is needed here (unlike `autosync-ic-skills`); the script parses
+versions with POSIX tools only.
+
+## Step 3 — Download the script
+
+The script is published as a file alongside this skill, so fetch it verbatim rather
+than transcribing it — that guarantees byte-exact content and keeps the channel
+detection correct as it is updated upstream:
+
+```bash
+mkdir -p .claude
+curl -fsSL https://skills.internetcomputer.org/.well-known/skills/autoupgrade-icp-cli/scripts/upgrade-icp-cli.sh \
+  -o .claude/upgrade-icp-cli.sh
+```
+
+Do **not** hand-write or paraphrase it.
+
+## Step 4 — Register the SessionStart hook (idempotently)
+
+Add a `SessionStart` hook to `.claude/settings.json`.
+
+- If `.claude/settings.json` does **not** exist, create it with the content below.
+- If it **does** exist, **merge** — preserve all existing keys, hooks, and permissions.
+  Only add the `SessionStart` entry, and **only if an equivalent
+  `bash .claude/upgrade-icp-cli.sh` command is not already present** (do not create a
+  duplicate). Parse the existing JSON, insert into the `hooks.SessionStart` array, and
+  write it back; never blindly overwrite the file.
+
+**Notify mode** (the default):
+
+```json
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          { "type": "command", "command": "bash .claude/upgrade-icp-cli.sh --mode notify" }
+        ]
+      }
+    ]
+  }
+}
+```
+
+**Auto mode** — same entry with `--mode auto`, plus a raised `timeout`. Installing a
+CLI binary can take well over the default 60 s hook timeout on a slow connection, and
+a timed-out upgrade leaves the tool half-installed:
+
+```json
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          { "type": "command", "command": "bash .claude/upgrade-icp-cli.sh --mode auto", "timeout": 300 }
+        ]
+      }
+    ]
+  }
+}
+```
+
+Switching modes later is just editing that `--mode` flag (and adding or dropping the
+`timeout`).
+
+Whenever you hand someone an auto-mode hook, say the thing that makes it a real
+choice: it upgrades the build toolchain **while a project is open**, so a session can
+start against a different `icp` than the last one ended with. That is exactly what
+some teams want and exactly what others pin against.
+
+## Step 5 — Run it once now
+
+```bash
+bash .claude/upgrade-icp-cli.sh --mode notify   # or --mode auto, matching the hook
+```
+
+Run it with the same mode you registered, so the user sees exactly what the hook will
+do. When both tools are current the script prints nothing and exits 0 — silence is the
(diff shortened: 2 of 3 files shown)


raymondk:
## Trimmed SKILL.md: 2,452 → 1,901 tokens (−22%), 1,888 → 1,271 words (−33%)

Organizing principle: the script is downloaded verbatim, never transcribed, so prose describing its internals was dead weight in SKILL.md — it already lives in the script's own header comments.

**Removed:** the `Behaviour worth knowing` section (JSON-vs-stdout hook mechanics, the exit-0 policy, the pre-release comparison rule), the table's "latest version from" column along with the Homebrew-lag and rate-limit rationale, and the "no `jq` needed" aside. Merged the run/verify/report steps; compressed the built-in-update-check comparison into the intro.

**Deliberately kept:** the channel table, the `ic-wasm` pairing rationale, the PATH-shadowing note, and the mode question — the evals show those four carrying the delta, so cutting them would cost capability rather than tokens. Both hook JSON blocks also stay verbatim: collapsing the auto one into a described delta saves ~120 tokens, but a malformed `settings.json` merge is a worse failure than a longer file.

### The re-run caught two real regressions, both fixed

1. **Table placeholders.** `<pkg>` / `<formula>` left the agent correctly naming `ic-wasm` as the culprit but never producing an upgrade command (eval 4: 3/3 → 2/3). Made literal per the repo writing guide — back to 3/3.
2. **Lost consequence.** The rewrite kept "a session can begin against a different `icp`" but dropped *why that matters* (eval 2: 4/4 → 3/4). Restoring "can change build behaviour and output" — back to 4/4.

In both cases the fix was the skill, not the assertion.

### Results after the trim

| Case | Before | After |
|---|---|---|
| 1 Setup plan and settings.json entry | 6/6 vs 2/6 | 6/6 vs 3/6 |
| 2 Auto mode needs a raised hook timeout | 4/4 vs 1/4 | 4/4 vs 0/4 |
| 3 Channel-correct upgrade command | 3/3 vs 2/3 | 3/3 vs 1/3 |
| 4 ic-wasm version skew | 3/3 vs 1/3 | 3/3 vs 1/3 |
| 5 Upgrade ran but version did not move | 4/4 vs 3/4 | 4/4 vs 4/4 |
| Triggers | 18/18 | 18/18 |

`skill-validator score` re-run with `--rescore` (the first attempt was a cache hit on the pre-trim file): still **4.50/5** with identical sub-scores, but the judge's comment shifted from *"the explanatory 'why' prose is useful but adds some tokens beyond the pure step list"* to crediting *"genuinely non-obvious operational detail … rather than generic CLI knowledge."*

### Two caveats on the numbers, for the reviewer

- **Case 5 is now non-discriminating** (4/4 both ways — the baseline diagnoses PATH shadowing unaided). It still works as a regression net, but it should not be read as evidence of a delta.
- **Case 2's baseline column is noisy.** Across four runs the no-skill agent scored 1/4, 1/4, 1/4 and 0/4, usually by stopping at a file-access request rather than answering; one full-suite run also hit the 120 s `ETIMEDOUT` and scored 0/4 on *both* configs, which is an infrastructure failure rather than a measurement. The with-skill 4/4 is the reproducible half.

<details>
<summary>Full post-trim suite output</summary>

```
Evaluating skill: autoupgrade-icp-cli
Output cases: Setup plan and settings.json entry, Auto mode needs a raised hook timeout, Channel-correct upgrade command, ic-wasm version skew after upgrading icp, Upgrade ran but the version did not move

━━━ Setup plan and settings.json entry ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 6/6 passed
    ✅ Asks the user whether the hook should only notify or should upgrade automatically
    ✅ Registers a SessionStart hook whose command runs the script
    ✅ Says the script is downloaded from the specified URL rather than hand-written
    ✅ Says existing settings.json must be merged, preserving other keys, and entry not duplicated/overwritten
    ✅ Mentions user will be asked to approve/trust the new hook before it runs automatically
    ✅ Does not invent a settings.json schema — nests under hooks.SessionStart with hooks array of {type, command}

  WITHOUT skill: 3/6 passed
    ❌ Asks the user whether the hook should only notify or upgrade automatically
       → The output silently designs a notify-only warning script and never offers or asks about an auto-upgrade option.
    ✅ Registers a SessionStart hook whose command runs the script
    ❌ Says the script is downloaded from the skills.internetcomputer.org URL rather than hand-written
       → The output explicitly instructs writing a new hand-authored checker script and never mentions downloading from that URL.
    ✅ Says existing settings.json must be merged, preserving other keys/hooks, without duplication or overwrite
    ❌ Mentions the user will be asked to approve/trust the new hook before it runs
       → There is no mention of a trust/approval prompt for the new hook anywhere in the output.
    ✅ Does not invent a settings.json schema

━━━ Auto mode needs a raised hook timeout ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 0/4 passed
    ❌ Uses the auto mode flag on the command (e.g. `--mode auto`)
       → No output was produced at all due to a spawn timeout error.
    ❌ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
       → No hook JSON was generated because the process errored out before producing any content.
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → No explanatory text was returned; the only output is an ETIMEDOUT error.
    ❌ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs
       → There is no claim of any kind since no substantive output was produced, so this behavior cannot be verified as present.

  WITHOUT skill: 0/4 passed
    ❌ Uses the auto mode flag on the command (e.g. `--mode auto`)
       → No hook JSON or command was produced at all; the assistant only asked for file access or a pasted config.
    ❌ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
       → No timeout value or hook JSON was given since the assistant deferred the task entirely.
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → The output contains no discussion of risks at all, only a request for more information.
    ❌ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs
       → While the assistant made no false background claim, it also never delivered any hook or explanation, so this required content is absent rather than correctly satisfied.

━━━ Channel-correct upgrade command ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 3/3 passed
    ✅ Gives `icp-cli-update` (the updater the shell installer places on PATH) as the upgrade command
    ✅ Does NOT tell the user to run `npm install -g @icp-sdk/icp-cli@latest`, which would install a second copy from a different channel
    ✅ Explains that the upgrade command depends on the channel the CLI was installed from

  WITHOUT skill: 1/3 passed
    ❌ Gives `icp-cli-update` (the updater the shell installer places on PATH) as the upgrade command
       → The output recommends `icp upgrade` instead of `icp-cli-update`.
    ✅ Does NOT tell the user to run `npm install -g @icp-sdk/icp-cli@latest`, which would install a second copy from a different channel
    ❌ Explains that the upgrade command depends on the channel the CLI was installed from
       → The explanation asserts a single built-in self-update command without noting that the correct command varies by install channel.

━━━ ic-wasm version skew after upgrading icp ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 2/3 passed
    ✅ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
    ❌ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
       → It tells the user to run `ic-wasm --version` to compare versions but never gives an actual upgrade command or instruction.
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

  WITHOUT skill: 1/3 passed
    ❌ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
       → The output only vaguely references ic-wasm as an example of a wasm-opt wrapper in item 3, but never identifies ic-wasm specifically as the tool that lags behind icp version bumps.
    ❌ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
       → There is no mention of checking `ic-wasm --version` or any specific upgrade command for ic-wasm.
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

━━━ Upgrade ran but the version did not move ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 4/4 passed
    ✅ Identifies that a second copy of icp from a different install channel is earlier on PATH and shadowing the npm one (commonly ~/.cargo/bin/icp from the shell installer)
    ✅ Suggests resolving the actual binary being used, e.g. `command -v icp` / `which icp` (and following the symlink) to see which copy wins
    ✅ Says upgrading through one channel will never move a copy installed through another — remove one, or upgrade the one that is actually on PATH
    ✅ Does NOT blame the npm cache or recommend reinstalling npm as the primary explanation

  WITHOUT skill: 4/4 passed
    ✅ Identifies that a second copy of icp from a different install channel is earlier on PATH and shadowing the npm one (commonly ~/.cargo/bin/icp from the shell installer)
    ✅ Suggests resolving the actual binary being used, e.g. `command -v icp` / `which icp` (and following the symlink) to see which copy wins
    ✅ Says upgrading through one channel will never move a copy installed through another - remove one, or upgrade the one that is actually on PATH
    ✅ Does NOT blame the npm cache or recommend reinstalling npm as the primary explanation

━━━ Trigger Evals ━━━

  Running trigger evaluation...

  Should trigger: 9/9 correct
    ✅ "Can you make this project always use the latest icp CLI?"
    ✅ "Set up a hook so I stop working with an outdated icp version"
    ✅ "I keep forgetting to update icp-cli. Automate it for me."
    ✅ "How do I upgrade the ICP CLI to the newest release?"
    ✅ "My icp is on 1.2.0 and the latest is 1.5.0 — what's the upgrade command?"
    ✅ "Install something that keeps icp and ic-wasm current at the start of every session"
    ✅ "I ran npm install -g @icp-sdk/icp-cli@latest but icp --version hasn't changed"
    ✅ "Is my ic-wasm too old for the version of icp I'm running?"
    ✅ "Add automatic ICP toolchain updates to .claude/settings.json"

  Should NOT trigger: 9/9 correct
    ✅ "Upgrade my canister without wiping its state — I think I need icp deploy --mode upgrade"
    ✅ "Bump the @dfinity/motoko recipe in my icp.yaml to the latest version"
    ✅ "Keep my Internet Computer skills in .claude/skills up to date automatically"
    ✅ "Update the moc compiler version in my mops.toml toolchain section"
    ✅ "My icp deploy is failing with a build error in the rust recipe"
    ✅ "What's the latest version of @icp-sdk/core I should depend on in my frontend?"
    ✅ "Write a GitHub Actions workflow that deploys my canister on every push to main"
    ✅ "How do I upgrade a canister's Wasm module on mainnet without losing stable memory?"
    ✅ "Add a SessionStart hook that runs my test suite before I start working"

━━━ Summary ━━━

  Output evals:
    Setup plan and settings.json entry: WITH 6/6 | WITHOUT 3/6
    Auto mode needs a raised hook timeout: WITH 0/4 | WITHOUT 0/4
    Channel-correct upgrade command: WITH 3/3 | WITHOUT 1/3
    ic-wasm version skew after upgrading icp: WITH 2/3 | WITHOUT 1/3
    Upgrade ran but the version did not move: WITH 4/4 | WITHOUT 4/4
  Trigger evals: should-trigger 9/9 | should-not-trigger 9/9

Full results saved to: /home/raymond/dev/icskills/evaluations/results/autoupgrade-icp-cli-2026-09-11T16-39-21-532Z.json
```

</details>

<details>
<summary>Eval 4 re-run (literal commands) and eval 2 re-run (restored consequence)</summary>

```
Evaluating skill: autoupgrade-icp-cli
Output cases: ic-wasm version skew after upgrading icp

━━━ ic-wasm version skew after upgrading icp ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 3/3 passed
    ✅ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
    ✅ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

  WITHOUT skill: 1/3 passed
    ❌ Points at ic-wasm being left behind while icp moved forward, since the official recipes invoke ic-wasm during every build
       → The response never mentions ic-wasm at all, focusing instead on moc, changelogs, npm install state, and version pins.
    ❌ Tells the user to check `ic-wasm --version` and upgrade it (e.g. `npm install -g @icp-sdk/ic-wasm@latest`)
       → No command or instruction to check or upgrade ic-wasm is given anywhere in the response.
    ✅ Does NOT send the user straight into debugging their Motoko code or recipe config as the first step

━━━ Summary ━━━

  Output evals:
    ic-wasm version skew after upgrading icp: WITH 3/3 | WITHOUT 1/3

Full results saved to: /home/raymond/dev/icskills/evaluations/results/autoupgrade-icp-cli-2026-09-11T16-41-09-019Z.json

Evaluating skill: autoupgrade-icp-cli
Output cases: Auto mode needs a raised hook timeout

━━━ Auto mode needs a raised hook timeout ━━━

  Running WITH skill...
  Running WITHOUT skill...
  Judging WITH skill...
  Judging WITHOUT skill...

  WITH skill: 4/4 passed
    ✅ Uses the auto mode flag on the command (e.g. `--mode auto`)
    ✅ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
    ✅ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
    ✅ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs

  WITHOUT skill: 0/4 passed
    ❌ Uses the auto mode flag on the command (e.g. `--mode auto`)
       → The output contains no hook JSON or command at all, only a permission request.
    ❌ Raises the hook timeout above the 60 second default (e.g. "timeout": 300)
       → No hook configuration was produced in the output.
    ❌ Explains the risk of auto mode: a CLI upgrade can change build behaviour mid-project
       → No risk explanation was given since the task stalled at a permission request.
    ❌ Does NOT claim the upgrade runs in the background — the hook blocks session start while it runs
       → While no false claim was made, this is only because no substantive response was produced, so the behavior is not demonstrated as present.

━━━ Summary ━━━

  Output evals:
    Auto mode needs a raised hook timeout: WITH 4/4 | WITHOUT 0/4

Full results saved to: /home/raymond/dev/icskills/evaluations/results/autoupgrade-icp-cli-2026-09-11T16-44-13-304Z.json
```

</details>

